PR #47, opened 2018-07-28
A fix for AVR-112?

Dusk-Argentum:
Please note, since running Avrae takes way too many steps, I opted out of doing a livetest. If this "fix" breaks more than it fixes, feel free to yell at me and decline the PR.

Also, maybe add "how to run Avrae locally" to the readme?


Dusk-Argentum:
Also, if this doesn't fix what it's meant to, or AVR-112 isn't *technically* a bug, feel free to demolish me with your eye lasers. Thanks!

mommothazaz123:
This change would break the aliasing system.

PR #71, opened 2018-08-02
Allow resist args for `!i opt` be case insensitive

Croebh:
Simply .lower() the input of args for `-resist` `-immune` and `-vuln`
So that 
Bludgeoning == bludgeoning == BluDgeoNing
Biggest issue I see is any potential resists that exist before this that have capitals can't be removed? I wasn't sure how to fix that, other than something like

combatant.resists['resist'] = list(set(i.lower() for i in combatant.resists['resist']))

But I figured you wouldn't want that running every time opt? I dunno


mommothazaz123:
(Tagging #73)

PR #75, opened 2018-08-03
`-miss` arg for attacks

Croebh:
Probably wrong, but I feel like I'm learning at the very least?

### DIFF
--- a/cogs5e/funcs/sheetFuncs.py
+++ b/cogs5e/funcs/sheetFuncs.py
@@ -119,11 +119,11 @@ def sheet_attack(attack, args, embed=None):
         else:
             if args.get('hit'):
                 out += "**To Hit**: Automatic hit!\n"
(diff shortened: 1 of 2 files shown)


cogs5e/funcs/sheetFuncs.py
```diff
@@ -120,6 +121,9 @@ def sheet_attack(attack, args, embed=None):
                 out += "**To Hit**: Automatic hit!\n"
             if args.get('crit'):
                 itercrit = args.get('crit', 0)
+            if args.get('miss'):
```
mommothazaz123: This should probably be right under the check for hit, and be elif

mommothazaz123:
(tagging #73)

Croebh:
What would I need to change for init?

mommothazaz123:
Actually nothing, apparently I didn't remember correctly.

PR #77, opened 2018-08-03
Limit multiline to defined number of commands

nadizak:
Using loops you can get multiline to run for a very long time.
Limiting it to 100 helps fix potential spam abuse.

Maybe lower the number to about 40-60 instead. (Meaning avrae will only run commands for a full minute or so and will stop after that)


nadizak:
For reference to why we should have a limit if you didn't see it in offtopic:

I made an alias print the entire script of the Bee Movie. It took 27 minutes of echoing once per second to finish.

Imagine that in an actual server...

PR #100, opened 2018-08-09
Allow `!multiline` to have `\n`

Croebh:


cogsmisc/customization.py
```diff
@@ -76,20 +76,15 @@ def __init__(self, bot):
...
             guild_prefix = self.bot.prefix
-        cmds = cmds.splitlines()
-        output = []
-        for cmd in cmds:
-            if cmd.startswith(guild_prefix) or cmd.startswith(self.bot.prefix) or not output:
-                output.append(cmd)
-            else:
-                output[-1] += "\n" + cmd
-        cmds = output
+        cmds = cmds.split(guild_prefix)
```
mommothazaz123: This will cause unintended behaviour if the prefix is found anywhere in the command (e.g. `!`)

PR #114, opened 2018-08-16
Multiline Change for Alternate Prefix Servers

nadizak:
I dunno if this change breaks any intended fail safes.
Since multiline is processed in bulk on the backend, allowing the commands in multiline to use ! as opposed to
the guild prefix shouldn't cause any issues with other bots.

Trying to think of what the intent was behind killing the multiline command if it used the default prefix?


nadizak:
AFR-146 I guess

PR #121, opened 2018-08-21
File and Gvar alias creation

nadizak:
Let's you create alias and server alias using either a plain text file or a gvar.
File upload limited to 4000 chars for aliases.
example use:
!alias tast file 
  (with a simple text file upload)
!servalias tast gvar <name-of-gvar>


mommothazaz123:
I know you're already working on making gvar aliases dynamic - another change to keep the codebase neat would be making `!alias file [name]` and `!alias gvar [name] [gvar]` subcommands of alias instead of a check in the main alias command.

nadizak:
Made a small change to fix a bug a ddb user was having. Dunno how you can pull just that one but there it is haha.

mommothazaz123:
It'd actually be easier to just check `if not attack.get('details')`

nadizak:
That's what I was going to do but figured you might've had a reason to check only for None. :man_shrugging:

nadizak:
Zhu I did the thing. >_>

Made them subcommands as well as just wrapping a generic load_file_for_alias bit.

Removed the gvar option for now.

Limit still 4000 and I added a bit for truncated display of alias when finished. Not sure on the number though (1500). Goal for it is to just cut it off and display . . . . .  for the rest.

nadizak:
It should work just fine but i'm going to change the display of the alias after it is loaded and how it is displayed. (Though i don't think anyone will be making any that would be more than the limit anytime soon)

mommothazaz123:
Can you reopen this as a new PR? It's hard to see exactly what you changed with 120 of my commits mixed in.

PR #198, opened 2018-10-11
Clarify error

Croebh:


mommothazaz123:
Don't think this particular instance requires further clarification.

PR #206, opened 2018-10-15
Alias from File

nadizak:
400 char limit
alias and servalias

dooink

### DIFF
--- a/cogsmisc/customization.py
+++ b/cogsmisc/customization.py
@@ -27,6 +27,7 @@ from cogs5e.models.errors import NoCharacter, EvaluationError, FunctionRequiresC
 from utils.functions import confirm, clean_content
 
 MAX_FILE_ALIAS_SIZE = 4000
+ALIAS_DISPLAY_LIMIT = 1500
 
 class Customization:
     """Commands to help streamline using the bot."""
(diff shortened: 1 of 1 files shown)


cogsmisc/customization.py
```diff
@@ -145,6 +149,20 @@ def handle_alias_arguments(self, command, message):
...
+            return await self.bot.say('You must supply an alias name.')
+        if not 0 < len(ctx.message.attachments) < 2:
+            return await self.bot.say('You must attach a plain text file to use a file as the alias source.')
+        file = ctx.message.attachments[0]
+        async with aiohttp.get(file['url']) as f:
+          cmds = await f.read()
+          if len(cmds) > MAX_FILE_ALIAS_SIZE:
+              return await self.bot.say('Alias file upload is limited to {} characters.'.format(MAX_FILE_ALIAS_SIZE))
+        cmds = cmds.decode("utf-8")
+        return await ctx.invoke(self.bot.get_command("servalias" if servalias else "alias"), alias_name=alias_name, cmds=cmds)
```
mommothazaz123: This seems a little janky; why not make the routine just get the contents of the file, and set in the invoking command?

nadizak:
It's all functional. Not sure on the alias truncation print. (Right now it's just a bunch of dots)

cogsmisc/customization.py
nadizak: set in the invoking command?

nadizak: Ah okay i see what you mean. Yeah I can change it to that I guess. I missed a ( on the truncation string anyway so i'll do another commit later to fix those.

PR #220, opened 2018-10-17
Possibly allow -f in `!i cast`

Croebh:


mommothazaz123:
The function is not imported and will crash.

PR #242, opened 2018-10-26
Change Critical to 25%

Croebh:
Silver made me do it, he's too lazy to github himself.


mommothazaz123:
I chose 15% since it's the threshold where 1-2 hits will kill the combatant, and is high enough that players are likely to see critical instead of just bloodied->dead; this threshold isn't going to change

PR #279, opened 2018-11-16
Allow Dicecloud to just add [^] to an attack

Croebh:


utils/functions.py
```diff
@@ -209,7 +209,7 @@ def parse_resistances(damage, resistances, immunities, vulnerabilities, neutral=
                 roll_string = '({0}) * 2'.format(roll_string)
                 checked.append(vulnerability)
                 break
-        if not comment.endswith('^'):
+        if not comment.endswith('^') or not comment.startswith('^ '):
```
mommothazaz123: `^ `? (note the space)

mommothazaz123:
```suggestion
        if not comment.endswith('^') or not comment.startswith('^'):
```
I prefer it less strict, so other sheet systems could use it too

PR #285, opened 2018-11-19
Allow `combatant.attacks` in scripting.

Croebh:


cogs5e/funcs/scripting.py
```diff
@@ -330,6 +330,7 @@ def __init__(self, combatant: Combatant, hidestats=True):
             self.initmod = None
             self.temphp = None
             self.resists = None
+        self.attacks = self._combatant.attacks
```
mommothazaz123: Should attacks only be visible for non-hidden combatants?

Croebh: I was unsure, but I guess not

PR #342, opened 2018-12-18
Fix 'RageDamage' in attacks

Croebh:


mommothazaz123:
I'd prefer if this calculated `rageDamage` and cached it in the evaluator instead of recalculating on each attack

cogs5e/sheets/dicecloud.py
```diff
@@ -442,6 +442,7 @@ def get_attacks(self):
         """Returns a list of dicts of all of the character's attacks."""
         if self.character is None: raise Exception('You must call get_character() first.')
         character = self.character
+        self.ragedmg = str(self.get_stat('rageDamage'))
```
mommothazaz123: ragedmg shouldn't be stored as an instance variable in the SheetParser object

Croebh: Would it still work if it was just `ragedmg = ` and calling `ragedmg` in the attack parser?

PR #343, opened 2018-12-18
Add a control character ^ to attacks for proper noun

Croebh:


utils/functions.py
```diff
@@ -166,6 +166,8 @@ def search(list_to_search: list, value, key, cutoff=5, return_key=False, strict=
 
 
 def a_or_an(string, upper=False):
+    if re.match('^', string):
+        return '{0}'.format(string.replace('^',''))
```
mommothazaz123: `re.match`: more expensive; use `string.startswith` instead
format: simply do `return string.lstrip`, no need for the unnecessary format

mommothazaz123:
Resolves #343

mommothazaz123:
Er. #343 is the PR itself, I mean #132

PR #351, opened 2018-12-20
Fix for Conc overriding spell

Croebh:


Croebh:
It was this or go through every single spell and come up with an alternative name for every effect :-/

mommothazaz123:
Please do the latter.

PR #365, opened 2018-12-26
change get_mod to use bitshifts

ShunyaoLiang:
Bitshifting (should be) faster than flooring a division result.

### DIFF
--- a/cogs5e/models/monster.py
+++ b/cogs5e/models/monster.py
@@ -26,11 +26,15 @@ class AbilityScores:
         self.charisma = cha
 
     def get_mod(self, stat):
-        return {'str': floor(self.strength / 2 - 5), 'dex': floor(self.dexterity / 2 - 5),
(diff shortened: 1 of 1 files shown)


mommothazaz123:
There appears to be no notable improvement.
```py
>>> timeit.timeit("[astom(i) for i in range(21)]", globals={'astom': astom})
=5.600126069970429 us/loop
```
```py
>>> timeit.timeit("[floor(i/2-5) for i in range(21)]", globals={'floor': floor})
=4.750285131391138 us/loop
```
Tested with 1,000,000 loops for each.

64:
In fact it's only slower because of the function call overhead.

```python
>>> from math import floor
>>> timeit.timeit("[floor(i/2-5) for i in range(21)]", globals={'floor': floor})
5.599004378000004
>>> timeit.timeit("[i - 10 >> 1 for i in range(21)]")
2.1754564479999914
```

However, there's no point merging the change without the function IMO. The readability that the function adds is the only reason to merge something like this; a ~20ns improvement is totally irrelevant.

PR #410, opened 2019-01-15
Add `!feat add [featname]`

Croebh:


mommothazaz123:
It feels a bit unintuitive for this to be in the lookup cog

PR #418, opened 2019-01-16
AUT-007 User Automation: 'Guidance' by tiberius0954

Croebh:
Removes all dead monster combatants.


mommothazaz123:
...? This is already automatically done by the tracker.

PR #577, opened 2019-04-16
Adds a `-ping` arg to both `!i opt`

Croebh:
Toggles whether the combatants controller gets pinged when attacked.


mommothazaz123:
After some discussion with the author of this PR, we've concluded that this change would break pre-existing combats and would add a required positional arg to the model for what's essentially a cosmetic change (c.f. the standard dictionary for cosmetic changes). I'm going to close this for now, but am open to revisiting a solution to this in the future.

PR #594, opened 2019-05-04
Slightly improve guide for local copy

Croebh:
Also edit the requirements to use a specific version (that runs fine) of discord.py, rather than the most recent.


mommothazaz123:
Closing as this is outdated with #631 and #621

PR #606, opened 2019-05-19
subscriptable SimpleCombatant

nadizak:
Only let's you subscript attributes.
Will throw attribute error on funcs and "class" variables

I got bored and tried to mess with the !browse alias

Didn't know if you would consider it all right to access funcs and hidden stuff with combatant[string] but i figured having at least stuff like combatant["hp"] work would be fine.

Only tested it via manually creating a simple combatant cause i don't feel like trying to get the whole thing to start

[test.txt](https://github.com/avrae/avrae/files/3194750/test.txt)
(Renamed it cause i guess they don't let you attach python files to these...)

### DIFF
--- a/cogs5e/funcs/scripting/combat.py
+++ b/cogs5e/funcs/scripting/combat.py
@@ -203,6 +203,12 @@ class SimpleCombatant:
     def __str__(self):
         return str(self._combatant)
 
+    def __getitem__(self, item):
+        attr = getattr(self, item)
+        if callable(attr) or item.startswith("_"):
+            raise AttributeError
+        return attr
+
 
 class SimpleGroup:
     def __init__(self, group: CombatantGroup):


mommothazaz123:
What exactly is the use case of this, outside of the browse alias?

nadizak:
To be honest I do not know. I haven't really touched combat aliasing. Would maybe be a better question for Deri Croebh and Silver xD

It would change this to like less than 10% of it via gvars though. 
```{x.name:{"name":x.name,"ac":x.ac,"hp":x.hp,"init":x.init,"initmod":x.initmod,"maxhp":x.maxhp,"temphp":x.temphp,"note":x.note,"ratio":x.ratio,"hpstr":x.hp_str(),"type":x.type,"level":x.level,"effects":{ y.name:{"name":y.name,"duration":y.duration,"remaining":y.remaining,"conc":y.conc,"effect":y.effect} for y in x.effects},"attacks":x.attacks}for x in combat().combatants}```

PR #621, opened 2019-06-12
DDBADB-3 Docker support

madcowfred:
- Add a Dockerfile
- Add Docker Compose config to create and run bot/Mongo/Redis instances
- Change static JSON loading to default to empty data structures if files aren't found
- Change various cases of hardcoded credential usage to use environment variables instead


docker/credentials-production.py
```diff
@@ -0,0 +1,32 @@
...
+mongo_url = os.environ['MONGO_URL']
+
+# The URI of a Redis cache (e.g. redis://localhost:6379/0)
+redis_url = os.environ['REDIS_URL']
+
+# A Dicecloud username.
+dicecloud_user = os.environ['DICECLOUD_USER']
+
+# The Dicecloud password of the Dicecloud user.
+dicecloud_pass = os.getenv['DICECLOUD_PASS']
```
mommothazaz123: oh, just realized - this isn't valid, is it?

madcowfred: Oops, fixed.

PR #632, opened 2019-06-17
Bugfixes: 2019w25

mommothazaz123:
Fixes:
- AVR-321 #615 - This is done by running `fuzz.partial_match` on `key(first_result)` and `query` - if they are less than 75% similar, it'll open a dialog instead of assuming the only match is correct
Example: `!feat` only returns Grappler, the only SRD feat
    - `!feat crossbow expert` opens dialog
    - `!feat grap` returns Grappler
    - `!feat gpler` returns Grappler
    - `!feat gour` opens dialog

- AVR-324 #627 

Other changes:
- Removed some old search methods
- Moved some functions over to using `search_and_select` instead of manually building select list and using `get_selection` (namely, monster attack and init attack)

### DIFF
--- a/utils/functions.py
+++ b/utils/functions.py
@@ -138,6 +138,9 @@ async def search_and_select(ctx, list_to_search: list, query, key, cutoff=5, ret
     if strict:
         result = results
     else:
+        if len(results) == 0:
+            raise NoSelectionElements()
+
         first_result = results[0]
         confidence = fuzz.partial_ratio(key(first_result).lower(), query.lower())
         if len(results) == 1 and confidence > 75:
--- a/utils/functions.py
+++ b/utils/functions.py
@@ -53,24 +53,6 @@ def get_positivity(string):
         return None
 
 
-def strict_search(list_to_search: list, key, value):
-    """Fuzzy searches a list for a dict with a key "key" of value "value" """
-    result = next((a for a in list_to_search if value.lower() == a.get(key, '').lower()), None)
-    return result
-
-
-def fuzzy_search(list_to_search: list, key, value):
-    """Fuzzy searches a list for a dict with a key "key" of value "value" """
-    try:
-        result = next(a for a in list_to_search if value.lower() == a.get(key, '').lower())
-    except StopIteration:
-        try:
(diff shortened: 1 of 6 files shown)


utils/functions.py
```diff
@@ -86,9 +68,9 @@ def search(list_to_search: list, value, key, cutoff=5, return_key=False, strict=
     if result is None:
         partial_matches = [a for a in list_to_search if value.lower() in key(a).lower()]
         if len(partial_matches) > 1 or not partial_matches:
-            names = [key(d) for d in list_to_search]
-            fuzzy_map = {key(d): d for d in list_to_search}
-            fuzzy_results = [r for r in process.extract(value, names, scorer=fuzz.ratio) if r[1] >= cutoff]
+            names = [key(d).lower() for d in list_to_search]
```
mommothazaz123: It turns out some scores returned by `fuzz.ratio` were lower than they should have been due to case sensitivity - this fixes that

utils/functions.py
```diff
@@ -115,13 +97,13 @@ def search(list_to_search: list, value, key, cutoff=5, return_key=False, strict=
         return result, True
 
 
-async def search_and_select(ctx, list_to_search: list, value, key, cutoff=5, return_key=False, pm=False, message=None,
+async def search_and_select(ctx, list_to_search: list, query, key, cutoff=5, return_key=False, pm=False, message=None,
```
mommothazaz123: renamed `value` to `query` to be more clear

utils/functions.py
```diff
@@ -277,17 +266,17 @@ def paginate(iterable, n, fillvalue=None):
     return [i for i in zip_longest(*args, fillvalue=fillvalue) if i is not None]
 
 
-async def get_selection(ctx, choices, delete=True, return_name=False, pm=False, message=None):
+async def get_selection(ctx, choices, delete=True, pm=False, message=None, force_select=False):
```
mommothazaz123: removed the `return_name` kwarg and added `force_select`
- the former was not used anywhere and not well-documented
- the latter forces a dialog, even if there's only one choice

PR #633, opened 2019-06-17
Added create_cc function to scripting

Infinidoge:
Adds a create_cc function which acts like create_cc_nx except in that if the cc already exists, it will be replaced and if it does not already exist, created.

### DIFF
--- a/cogs5e/funcs/scripting/evaluators.py
+++ b/cogs5e/funcs/scripting/evaluators.py
@@ -159,7 +159,7 @@ class ScriptingEvaluator(EvalWithCompoundTypes):
                 self.character_changed = True
 
         def create_cc(name: str, **kwargs):
-            if name in set(con.name for con in character.consumables):
+            if cc_exists(name):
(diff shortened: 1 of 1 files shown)


cogs5e/funcs/scripting/evaluators.py
```diff
@@ -158,6 +158,11 @@ def create_cc_nx(name: str, minVal: str = None, maxVal: str = None, reset: str =
                 character.consumables.append(new_consumable)
                 self.character_changed = True
 
+        def create_cc(name: str, **kwargs):
+            if name in set(con.name for con in character.consumables):
```
mommothazaz123: perhaps use `if cc_exists(name)`?

Infinidoge: Oh yeah, that certainly would be more elegant. Fixed in latest commit.
Also, shouldn't create_cc_nx use `not cc_exists(name)`?

PR #648, opened 2019-06-20
Adds `!tembed`, an embed version of `!test`

### LINKED ISSUE #647: AFR-371 An embed equivalent of `!test`, `!tembed`
Feature Request: An embed equivalent of `!test`, `!tembed`
Who would use it? Aliasers
What part of the bot? Customization/scripting
How would it work? Just like `!test`, but instead of an echo, an embed
Why should it be added? The only way to test scripting in an embed is to create an alias.

- Croebh#5603

> Upvote - Croebh#5603

> Upvote - Poliphas#3955

> Upvote - Caellorn#0718

> Upvote - Infinidoge#1337

> Upvote - Derixyleth#0636

> Upvote - Unaert#4496

> Upvote - ShadowTheIncubus#0835

> Upvote - CaptDagobeard#1803

Votes: +9 / -0

Croebh:
Resolves #647

### DIFF
--- a/cogsmisc/customization.py
+++ b/cogsmisc/customization.py
@@ -7,6 +7,7 @@ import asyncio
 import textwrap
 import traceback
 import uuid
+import shlex
 
 import discord
 from discord.ext import commands
@@ -17,7 +18,7 @@ from cogs5e.funcs.scripting import ScriptingEvaluator
 from cogs5e.models.character import Character
 from cogs5e.models.embeds import EmbedWithAuthor
 from cogs5e.models.errors import AvraeException, EvaluationError, NoCharacter
-from utils.argparser import argquote, argsplit
+from utils.argparser import argquote, argsplit, argparse
 from utils.functions import auth_and_chan, clean_content, confirm
 
 ALIASER_ROLES = ("server aliaser", "dragonspeaker")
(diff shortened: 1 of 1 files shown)


cogsmisc/customization.py
```diff
@@ -369,6 +370,58 @@ def can_edit_servaliases(ctx):
         parsed = await char.parse_cvars(teststr, ctx)
         parsed = clean_content(parsed, ctx)
         await ctx.send(f"{ctx.author.display_name}: {parsed}")
+                       
+    @commands.command()
+    async def tembed(self, ctx, *, teststr):
```
mommothazaz123: Instead of rewriting the embed code, what you can do is have it parse the input string as above (L369-370), then call the `!embed` command with the parsed string (See sheetManager L179 for an example on how to do this)

Croebh: I think I did it properly? I also removed the imported I added

cogsmisc/customization.py
```diff
@@ -369,6 +369,24 @@ def can_edit_servaliases(ctx):
...
+	    -footer [footer text]
+	    -f ["Field Title|Field Text"]
+	    -color [hex color]
+	    -t [timeout (0..600)]
+	    """
+
+        char = await Character.from_ctx(ctx)
+        parsed = await char.parse_cvars(teststr, ctx)
+
+        return await ctx.invoke(self.bot.get_command('embed'), args=parsed)
```
mommothazaz123: `self.bot.get_command("embed")` should always work, but it's probably a good idea to have a handler for if it returns `None`

PR #654, opened 2019-06-26
Add reliable talent as a csetting

Croebh:
Hope I did this right, my local copy isn't updated past the character rewrite.


Infinidoge:
Verified as working on a local copy with all of the latest commits

PR #667, opened 2019-07-02
[GH-660] Automated Testing Framework

### LINKED ISSUE #660: Investigate/create automatic command testing framework
We should have some way of being able to test bot commands and responses automatically.

mommothazaz123:
Resolves #660 

### Summary
This PR sets up the framework to do tests on all Avrae commands without ever having to connect to Discord. This is done by overriding discord.py's HTTP implementation with our own to intercept calls to the Discord API, and bypassing the websocket to send messages to the bot directly.

Tests are done intuitively, by sending a message to the bot and checking what it sends to discord in a self-contained test. For example:
```py
async def test_ping(avrae, dhttp):
    avrae.message("!ping")
    await dhttp.receive_message("Pong."}
```
This code will send the message `!ping` to Avrae, and ensure that the bot replies with `Pong` within 10 seconds. If the received message is incorrect or times out, the test will fail.

#### PR Type

- [ ] This PR fixes an issue.
#### Other


mommothazaz123:
Some more items to consider later when writing tests:
- How should we test commands that require data? Should a small subset of static data be loaded in when a testing session starts (maybe dependent on #682)?
- How should we test commands that require characters? My thought for this is either having certain tests depend on the character loading test (using pytest-dependencies) or by having a project fixture load a static character.

PR #673, opened 2019-07-08
Hide resources with a value of 0 in Class lookup

### LINKED ISSUE #672: AFR-383 Hide Spell Slots with 0 value in `!class` command
**Feature Request:** Hide Spell Slots with 0 value in `!class` command
**Who would use it?** Everyone who uses lookup
**What part of the bot?** Lookup
**How would it work?** Check if the value in the table column is 0, and don't display if so.
**Why should this be added?** Reduce length/remove extraneous information

- silverbass#2407

> Upvote - Croebh#5603

> Upvote - Derixyleth#0636

> Upvote - Caellorn#0718

> Upvote - Infinidoge#1337

> Upvote - CaptDagobeard#1803

Votes: +6 / -0

Croebh:
### Summary
Only displays class resources if the value isn't 0.
Resolves #672

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/lookup.py
+++ b/cogs5e/lookup.py
@@ -244,7 +244,8 @@ class Lookup(commands.Cog):
                     level_resources[table['colLabels'][i]] = parse_data_entry([col])
 
             for res_name, res_value in level_resources.items():
-                embed.add_field(name=res_name, value=res_value)
+                if res_value != '0':
+                    embed.add_field(name=res_name, value=res_value)
 
             for f in level_features:
                 text = parse_data_entry(f['entries'])


cogs5e/lookup.py
```diff
@@ -244,7 +244,8 @@ def __init__(self, bot):
                     level_resources[table['colLabels'][i]] = parse_data_entry([col])
 
             for res_name, res_value in level_resources.items():
-                embed.add_field(name=res_name, value=res_value)
+                if res_value != '0':
```
mommothazaz123: As far as I can tell this only happens with higher level spell slots and a monk's Ki at level 1, so this should be fine.

PR #674, opened 2019-07-08
Automate testing with Docker

mommothazaz123:
### Summary
This PR sets up automated testing and code coverage in Travis CI using the testing framework, Docker, and CodeCov.

Merging into the existing automated testing branch to resolve the Docker testing point.

#### PR Type

- [ ] This PR fixes an issue.
#### Other


docker-compose.test.yml
```diff
@@ -0,0 +1,45 @@
...
+  mongo:
+    image: mongo
+    restart: always
+    environment:
+      MONGO_INITDB_ROOT_USERNAME: root
+      MONGO_INITDB_ROOT_PASSWORD: topsecret
+    ports:
+      - 58017:27017
+
+  mongo-express:
```
madcowfred: The test environment probably doesn't need Mongo Express

Dockerfile.test
```diff
@@ -0,0 +1,21 @@
+FROM python:3.6-stretch
+
+ARG DBOT_ARGS
```
madcowfred: Do tests do something different if you pass in "test" vs passing in nothing? If not, this can be removed.

Dockerfile.test
```diff
@@ -0,0 +1,21 @@
...
+ARG ENVIRONMENT=production
+
+WORKDIR /app
+
+COPY requirements.txt .
+RUN pip install -r requirements.txt
+
+RUN mkdir temp
+RUN mkdir shared
```
madcowfred: Minor: `RUN mkdir temp && mkdir shared` or `RUN mkdir temp shared` will generate one less interim Docker image as they are a single atomic unit of work.

Dockerfile.test
mommothazaz123: Some Avrae files check whether an environment is testing by checking for the presence of `"test" in sys.argv`, which this satisfies - it can probably be hardcoded for the tests, though

PR #682, opened 2019-07-10
✨ Mongo compendium data, automatic reloading

madcowfred:
### Summary
Use MongoDB as the source of compendium data

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/funcs/lookupFuncs.py
+++ b/cogs5e/funcs/lookupFuncs.py
@@ -3,6 +3,7 @@ Created on Jan 13, 2017
 
 @author: andrew
 """
+import asyncio
 import itertools
 import json
 import logging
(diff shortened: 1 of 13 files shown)


cogs5e/funcs/lookupFuncs.py
```diff
@@ -25,31 +26,98 @@
...
+        self.srd_races = self.read_json('srd-races.json', [])
+        self.srd_spells = self.read_json('srd-spells.json', [])
+
+        # Dictionary!
+        self.itemprops = self.read_json('itemprops.json', {})
+
+        self.load_common()
+
+    async def load_all_mongodb(self, mdb):
+        lookup = {d.key: d.object for d in await mdb.static_data.find({}).to_list(length=None)}
```
mommothazaz123: This should probably be `d['key']: d['object']`

madcowfred: Yup, found that out when I actually imported some data. Oops.

cogs5e/funcs/lookupFuncs.py
```diff
@@ -25,31 +26,98 @@
...
+
+    async def reload_task(self, mdb=None):
+        wait_for = int(os.getenv('RELOAD_INTERVAL', '300'))  # TODO: decide if 5 minutes is a reasonable default
+        if wait_for > 0:
+            log.info("Reloading data every {} seconds", wait_for)
+            while True:
+                await self.reload(mdb)
+                await asyncio.sleep(wait_for)
+
+    async def reload(self, mdb=None):
```
mommothazaz123: Just declaring the reload function async doesn't actually make it run in an executor by default - to do that, you'd want to make all the blocking calls a sync function (specifically L64-77 and `load_common()`), then call `loop.run_in_executor()`.

Check out utils/functions.py L500 (`generate_token()`) for an example

madcowfred: I miss so many Github notifications, oops.

The reload function should only ever be run in an executor, I... think? It's called from reload_task() which is started in bot.py as `bot.loop.create_task(compendium.reload_task(bot.mdb))`, and also called from a bot command. If there's extra work to do then Python async is even weirder than I thought!

madcowfred: Oh dear, it is weirder than I thought.

cogsmisc/adminUtils.py
```diff
@@ -155,6 +156,12 @@ def __init__(self, bot):
         await self.bot.change_presence(status=status, activity=discord.Game(msg or "D&D 5e | !help"))
         await ctx.send("Changed presence.")
 
+    @commands.command(hidden=True)
+    @checks.is_owner()
+    async def reload_static(self, ctx):
+        await compendium.reload(ctx.mdb)
```
mommothazaz123: should probably be `self.bot.mdb`

PR #684, opened 2019-07-11
[GH-676] Better Bestiary Storage

### LINKED ISSUE #676: Improve Bestiary Storage
Currently, the largest collection in Avrae's database is `bestiaries`, comprising over 50% of the database. This appears to be because there are *many* duplicate bestiaries imported, all of which are stored separately.

My proposed solution to this is to implement a sort of subscription system - if two users import the same bestiary, only the first import will be stored on the database, along with a hash. If an imported bestiary's hash is equal to one already stored in the database, the new user will be added as a subscriber to the existing bestiary, rather than storing another copy.

This allows for storing multiple versions of the same bestiary mastered on CritterDB.

mommothazaz123:
### Summary
Resolves #676 

The largest source of duplicate data in bestiaries is multiple users who have imported the same bestiary - the old system would store a copy for each user, leading to a lot of identical data. Since bestiaries are read-only, this change implements a hashing system to track identical bestiaries. If two users import an identical bestiary, it will now mark two "subscribers" on the same hashed bestiary, removing duplicate entries.

Running a test on a local copy of the database, this reduces the space needed to store bestiaries by 79.7% (as of 44ab986).

#### PR Type

- [ ] This PR fixes an issue.
#### Other
- [x] This PR is a breaking change (e.g. methods or parameters removed/renamed) (**A migrator script is included and will require downtime to run**)
- N/A I have updated the documentation to reflect the changes.

### DIFF
--- a/cogs5e/models/homebrew/bestiary.py
+++ b/cogs5e/models/homebrew/bestiary.py
@@ -3,7 +3,7 @@ import logging
 
 import aiohttp
 
-from cogs5e.models.errors import ExternalImportError, NoActiveBrew
+from cogs5e.models.errors import ExternalImportError, NoActiveBrew, NotAllowed
 from cogs5e.models.monster import Monster
 from utils.functions import search_and_select
 
@@ -21,7 +21,7 @@ class Bestiary:
         # subscription data - only atomic writes
         self.subscribers = subscribers
         self.active = active
-        self.server_active = server_active
+        self.server_active = server_active  # [{"subscriber_id": string, "guild_id": string}, ...]
 
         # content
         self.name = name
@@ -30,6 +30,8 @@ class Bestiary:
 
     @classmethod
     def from_dict(cls, d):
+        if 'monsters' in d:
+            d['monsters'] = [Monster.from_bestiary(m) for m in d['monsters']]
         return cls(**d)
 
     @classmethod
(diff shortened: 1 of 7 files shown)


cogs5e/models/homebrew/bestiary.py
```diff
@@ -1,56 +1,129 @@
...
+            raise AttributeError("load_monsters() must be called before accessing bestiary monsters.")
+        return self._monsters
+
+    async def write_to_db(self, ctx):
+        """Writes a bestiary object to the database. Returns self."""
+        assert self._monsters is not None
+        data = {
+            "sha256": self.sha256, "upstream": self.upstream, "subscribers": [str(ctx.author.id)], "active": [],
+            "server_active": [], "name": self.name, "desc": self.desc, "monsters": [m.to_dict() for m in self._monsters]
+        }
```
madcowfred: I find this difficult to parse, one key:value pair per line is much easier for me at least.

migrators/bestiary_hasher.py
```diff
@@ -0,0 +1,141 @@
...
+                            server_active, bestiary['name'], monsters, bestiary.get('desc'))
+    return new_bestiary
+
+
+def to_dict(bestiary):
+    return {
+        "sha256": bestiary.sha256, "upstream": bestiary.upstream, "subscribers": bestiary.subscribers,
+        "active": bestiary.active, "server_active": bestiary.server_active, "name": bestiary.name,
+        "desc": bestiary.desc, "monsters": [m.to_dict() for m in bestiary._monsters]
+    }
```
madcowfred: One key:value pair per line again.

Issue #706, opened 2019-07-24
AVR-359 Importing a sheet with a note for a bonus (AC, Skill, etc) without number


Infinidoge:
**Describe the bug**
Importing a sheet with a note for a bonus (AC, Skill, etc) when it does not have a number/selection for that bonus errors with the following:
```Error loading character: unsupported operand type(s) for +=: 'int' and 'NoneType'```
~~Additionally, not having a background selected reports the same error.~~ (Checked and not having a background works properly)

**Severity**
Low (formatting issues, things that don't impact operation) / Medium (minor functional impact)

**Context**
!beyond <sheet following step 1 below>
!update (done with a sheet active following step 1 below)

**To Reproduce**
1. Have a D&DBeyond sheet with a note for a bonus (AC, Skill, etc) without the numerical bonus specified ~~or without a background.~~
2. Import sheet (`!beyond` or `!update`)
3. See error: 
```Error loading character: unsupported operand type(s) for +=: 'int' and 'NoneType'```

**Expected behavior**
Sheet imports correctly, 
no bonus is added since the numerical bonus doesn't exist,
~~doesn't apply any background bonuses since background isn't selected~~

**Screenshots**
Example of notes
![Example of Notes](https://user-images.githubusercontent.com/22727114/61754178-1d452280-ad80-11e9-8a3e-c1458d42155c.PNG)
Erroring on import from `!beyond`
![Beyond Import](https://user-images.githubusercontent.com/22727114/61754266-8036b980-ad80-11e9-97d2-fdd17b40206a.PNG)
Erroring on import for `!update`
![Update Import](https://user-images.githubusercontent.com/22727114/61754348-c8ee7280-ad80-11e9-8eed-7cf42b077074.PNG)

Infinidoge:
Currently doing further testing as this has not been consistently reproducible

PR #707, opened 2019-07-24
AVR-359 Fix import error of houserules w/o values

Infinidoge:
### Summary
Fixes AVR-359 (#706) where sheets error on import due to houserules not having a value

This has generally occurred when they have a note but without a value, which led to them being in the list of values to apply with a NoneType as their value, causing an error on import
(Note: As far as I could tell this does not occur when houseruling proficiency type, only bonuses.)

This PR fixes the issue by adding Try/Except blocks to the relevant sections where it can error.

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/sheets/beyond.py
+++ b/cogs5e/sheets/beyond.py
@@ -280,8 +280,11 @@ class BeyondSheetParser(SheetLoaderABC):
                     skills[skill_name] = Skill(charval['value'])
                     ignored_ids.add(charval['valueId'])  # this must be the final value so we stop looking
                 elif charval['typeId'] in {24, 25}:  # PROBABLY skill magic/misc bonus
-                    skills[skill_name].value += charval['value']
-                    skills[skill_name].bonus += charval['value']
+                    try:
+                        skills[skill_name].value += charval['value']
+                        skills[skill_name].bonus += charval['value']
+                    except TypeError:
+                        pass  # listed due to note but probably doesn't have a value, ignore it
                 elif charval['typeId'] == 26:  # proficiency stuff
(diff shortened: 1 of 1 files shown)


cogs5e/sheets/beyond.py
```diff
@@ -359,9 +362,15 @@ def get_ac(self):
             if val['typeId'] == 1:  # AC override
                 return val['value']
             elif val['typeId'] == 2:  # AC magic bonus
-                miscBonus += val['value']
+                try:
```
mommothazaz123: Rather than duplicating a try/except block here, it seems like it would be more elegant to just add `if val['value'] is None: continue` at the start

cogs5e/sheets/beyond.py
```diff
@@ -280,8 +280,11 @@ def get_skills_and_saves(self):
                     skills[skill_name] = Skill(charval['value'])
                     ignored_ids.add(charval['valueId'])  # this must be the final value so we stop looking
                 elif charval['typeId'] in {24, 25}:  # PROBABLY skill magic/misc bonus
-                    skills[skill_name].value += charval['value']
-                    skills[skill_name].bonus += charval['value']
+                    try:
```
mommothazaz123: similarly, it seems more elegant to LBYL here in the conditional - that would also catch the type 23 case

cogs5e/sheets/beyond.py
Infinidoge: Hadn't thought of that, good point.

Infinidoge: Newest commit uses `if value None: continue` rather than Try/Except

PR #835, opened 2019-10-03
Allow for `!init opt` to modify values using + or -

Croebh:
### Summary
Allow for `!init opt` to modify values using + or -, adding the ability to make modifications on the fly without having to provide a hard value. Also adds delta's to those values.

Affects: `-p`, `-ac`, `-max`, and `-hp`

image

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/initTracker.py
+++ b/cogs5e/initTracker.py
@@ -537,10 +537,11 @@ class InitTracker(commands.Cog):
 
         def mod_or_set(opt_name, old_value):
             new_value = args.last(opt_name, type_=int)
-            if old_value is None:
-                old_value = 0
+            temp_value = old_value
+            if temp_value is None:
+                temp_value = 0
             if args.last(opt_name).startswith(('-', '+')):
-                new_value = old_value + new_value
+                new_value = temp_value + new_value
             return new_value, old_value
 
         @option()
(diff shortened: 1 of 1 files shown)


cogs5e/initTracker.py
```diff
@@ -535,6 +535,13 @@ def wrapper(func):
 
             return wrapper
 
+        def mod_or_set(opt_name, old_value):
+            new_value = args.last(opt_name, type_=int)
+            if args.last(opt_name).startswith(('-', '+')):
+                new_value = old_value + args.last(opt_name, type_=int)
```
mommothazaz123:
```suggestion
                new_value = old_value + new_value
```
Just reduces the number of calls.

cogs5e/initTracker.py
```diff
@@ -537,10 +537,11 @@ def wrapper(func):
 
         def mod_or_set(opt_name, old_value):
             new_value = args.last(opt_name, type_=int)
+            if old_value is None:
+                old_value = 0
```
mommothazaz123: Nit: This would display, for example, `HP max set to 10 (was 0)` instead of `(was None)`.

PR #841, opened 2019-10-07
Adds a `-parent` argument for `!init effect`

Croebh:
### Summary

Adds a `-parent` argument for `!init effect`, allowing a parent effect to be set outside of scripting.
Also adds group targeting for the first target of `!init effect` and for `!init re`.

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/initTracker.py
+++ b/cogs5e/initTracker.py
@@ -814,7 +814,7 @@ class InitTracker(commands.Cog):
                                         concentration=conc, tick_on_end=end)
                 result = combatant.add_effect(effect_obj)
                 if parent:
-                    combatant.get_effect(effect_name).set_parent(parent)
+                    effect_obj.set_parent(parent)
                 out = f"Added effect {effect_name} to {combatant.name}."
                 if result['conc_conflict']:
                     conflicts = [e.name for e in result['conc_conflict']]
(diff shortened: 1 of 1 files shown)


cogs5e/initTracker.py
```diff
@@ -784,13 +785,8 @@ def mod_or_set(opt_name, old_value):
         args = argparse(args)
 
         targets = []
-        first_target = await combat.select_combatant(target_name)
-        if first_target is None:
-            await ctx.send("Combatant not found.")
```
Croebh: So, in my testing this line would never run. `NoSelectionElements` was always returned before it could reach the if statement.

cogs5e/initTracker.py
```diff
@@ -800,16 +796,26 @@ def mod_or_set(opt_name, old_value):
...
             else:
-                effectObj = Effect.new(combat, combatant, duration=duration, name=effect_name, effect_args=args,
-                                       concentration=conc, tick_on_end=end)
-                result = combatant.add_effect(effectObj)
-                out = "Added effect {} to {}.".format(effect_name, combatant.name)
+                effect_obj = Effect.new(combat, combatant, duration=duration, name=effect_name, effect_args=args,
+                                        concentration=conc, tick_on_end=end)
+                result = combatant.add_effect(effect_obj)
+                if parent:
+                    combatant.get_effect(effect_name).set_parent(parent)
```
mommothazaz123:
```suggestion
                    effect_obj.set_parent(parent)
```
You already have a pointer to the existing Effect object - you don't need to get another.

PR #889, opened 2019-11-05
Clustering

mommothazaz123:
### Summary
Adds the code necessary for running multiple clusters, including an IPC system using Redis Pub/Sub. Also moves a lot of env var accesses to a single config file, so it's easier to track all of those down.

Adds boto3 and aioredis as dependencies (for checking what ECS tasks are alive, and for async access of redis)

#### PR Type

- [ ] This PR fixes an issue.
#### Other
    - all redis database calls are now async


zmckinnon:
I added the python team in hopes that they can give some good feedback.

JayCarney:
> in hopes that they can give some good feedback.

Man, no pressure or antyhing

cogs5e/models/dicecloud/client.py
```diff
@@ -1,21 +1,19 @@
...
+from utils import config
 from .errors import InsertFailure, LoginFailure
 from .http import DicecloudHTTP
 
-TESTING = (os.environ.get("TESTING", False) or 'test' in sys.argv)
-NO_DICECLOUD = os.environ.get("NO_DICECLOUD", False)
-UNAME = os.getenv('DICECLOUD_USER', 'avrae') if not TESTING else credentials.test_dicecloud_user
-PWD = credentials.dicecloud_pass.encode() if not TESTING else credentials.test_dicecloud_pass.encode()
-API_KEY = credentials.dicecloud_token if not TESTING else credentials.test_dicecloud_token
+TESTING = config.TESTING
```
madcowfred: Bit of a code smell here, these are purely aliasing config.FOO to FOO - probably better to remove these and replace the various FOO use points with config.FOO?

utils/redisIO.py
```diff
@@ -5,101 +5,190 @@
...
-        data = self.hget(key, field)
-        return json.loads(data) if data is not None else default
+    async def publish(self, channel, data):
+        return await self._db.publish(channel, data)
 
-    def jhset(self, key, field, value, **kwargs):
-        data = json.dumps(value, **kwargs)
-        return self.hset(key, field, data)
+
+class _PubSubMessageBase:
```
JayCarney: This should probably extend `abc.ABC` if it's intended to be abstract

utils/redisIO.py
```diff
@@ -5,101 +5,190 @@
...
+
+    def to_dict(self):
+        inst = super(PubSubCommand, self).to_dict()
+        inst.update({"command": self.command, "args": self.args, "kwargs": self.kwargs})
+        return inst
+
+
+class PubSubReply(_PubSubMessageBase):
+    def __init__(self, id, sender, reply_to, data):
+        super(PubSubReply, self).__init__('reply', id, sender)
```
JayCarney: python 3 doesn't require passing in the class for super()
```suggestion
        super().__init__('reply', id, sender)
```

utils/redisIO.py
```diff
@@ -5,101 +5,190 @@
...
+        self.reply_to = reply_to
+        self.data = data
+
+    @classmethod
+    def new(cls, bot, reply_to, data):
+        _id = str(uuid.uuid4())
+        return cls(_id, bot.cluster_id, reply_to, data)
+
+    def to_dict(self):
+        inst = super(PubSubReply, self).to_dict()
```
JayCarney:
```suggestion
        inst = super().to_dict()
```

utils/redisIO.py
```diff
@@ -5,101 +5,190 @@
...
+    def to_dict(self):
+        return {"type": self.type, "id": self.id, "sender": self.sender}
+
+    def to_json(self):
+        return json.dumps(self.to_dict())
+
+
+class PubSubCommand(_PubSubMessageBase):
+    def __init__(self, id, sender, command, args, kwargs):
+        super(PubSubCommand, self).__init__('cmd', id, sender)
```
JayCarney:
```suggestion
        super().__init__('cmd', id, sender)
```

JayCarney:
Random question, because I'm just missing some context; what leads to a ECS task not being alive and needing us to take over a dead cluster? just trying to understand the bigger picture.

mommothazaz123:
@JayCarney For context overall, Discord requires bots that are in over 2500 guilds (servers) to connect via shards, then sends all activity from one server to a given shard based on ``SHARD_ID = (GUILD_ID >> 13) % NUM_SHARDS``. If an ECS task dies (which has happened if discord sends bad data, or randomly disconnects us and the reconnect logic fails), the takeover code ensures that a new task runs the same shards as the one that died, so you don't have two clients responding to the same events.

JayCarney:
> @JayCarney For context overall, Discord requires bots that are in over 2500 guilds (servers) to connect via shards, then sends all activity from one server to a given shard based on `SHARD_ID = (GUILD_ID >> 13) % NUM_SHARDS`. If an ECS task dies (which has happened if discord sends bad data, or randomly disconnects us and the reconnect logic fails), the takeover code ensures that a new task runs the same shards as the one that died, so you don't have two clients responding to the same events.

Thanks for the context 👍

PR #942, opened 2019-12-18
Apply JoAT to ability checks.

thecoppersanctum:
### Summary
Updates the cell checked for JoAT since I moved it in 2.1. Then added a bit so it will apply the JoAT bonus, and also global "bonus to all ability checks" bonuses, from 2.0 onward (like Beyond does). And then changed a few == to >=

#### PR Type

- [ ] This PR fixes an issue.
- [x ] This PR adds a new feature that is not an open feature request.
#### Other

### DIFF
--- a/cogs5e/sheets/gsheet.py
+++ b/cogs5e/sheets/gsheet.py
@@ -130,7 +130,7 @@ class GoogleSheet(SheetLoaderABC):
     def __init__(self, url):
         super(GoogleSheet, self).__init__(url)
         self.additional = None
-        self.version = 1
+        self.version = (1, 0)  # major, minor
 
         self.total_level = 0
 
@@ -198,7 +198,9 @@ class GoogleSheet(SheetLoaderABC):
         doc = GoogleSheet.g_client.open_by_key(self.url)
         self.character_data = TempCharacter(doc.sheet1)
         vcell = self.character_data.value("AQ4")
-        if ("1.3" not in vcell) and vcell:
+        if '1.3' in vcell:
+            self.version = (1, 3)
+        elif vcell:
             self.additional = TempCharacter(doc.get_worksheet(1))
             self.version = (2, 1) if "2.1" in vcell else (2, 0) if "2" in vcell else (1, 0)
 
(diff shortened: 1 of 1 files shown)


cogs5e/sheets/gsheet.py
```diff
@@ -360,7 +360,7 @@ def get_skills_and_saves(self):
 
         skills = {}
         saves = {}
-        is_joat = self.version == 2 and bool(character.value("AR45"))
+        is_joat = self.version == 2 and bool(character.value("AR45")) or self.version == 2.1 and bool(character.value("AQ59"))
```
Croebh: I feel like this needs some parentheses for each statement on either side of the `or`

cogs5e/sheets/gsheet.py
```diff
@@ -369,12 +369,12 @@ def get_skills_and_saves(self):
             else:
                 profcell = None
             try:
-                value = int(character.value(cell))
+                value = int(character.value(cell)) + int("C" in cell and (self.version == 2.1 and character.value("AR58") or self.version == 2 and character.value("AQ26"))) + int("C" in cell and is_joat and int(character.value("H14"))//2)
```
Croebh: Why do the check again, instead of just reading `is_joat`?

thecoppersanctum: It does check `is_joat` where it needs to in there. The first plus pulls in the globally applied bonus.

cogs5e/sheets/gsheet.py
```diff
@@ -198,7 +198,7 @@ def _gchar(self):
         vcell = self.character_data.value("AQ4")
         if ("1.3" not in vcell) and vcell:
             self.additional = TempCharacter(doc.get_worksheet(1))
-            self.version = 2 if "2" in vcell else 1
+            self.version = 2.1 if "2.1" in vcell else 2 if "2" in vcell else 1
```
mommothazaz123: Generally, it's better to keep version numbers as a tuple of `(major, minor, patch)` (i.e. 2.1 should be version `(2, 1)`) - previously this didn't matter since there wasn't any difference in between minor versions

cogs5e/sheets/gsheet.py
mommothazaz123: Can you give an explanation of exactly what this is doing? Why do we need to check all these additional cells instead of just reading the final bonus presented on the sheet?

thecoppersanctum: This is taking the values of the ability modifiers (`"C"in cell`) which aren't a "final bonus" and adding the bonus that applies to all ability checks from things like a Stone of Good Luck (the value in AR58 for 2.1 or AQ26 for 2.0) and then adding the JoAT half proficiency bonus if applicable. This will make it consistent with how Beyond handles things and make it so you won't need a `joat` or `ls` snippet to apply Jack of All Trades or Luck Stone bonus to non-skill ability checks. Currently, everything but the straight ability checks include all this as presented from the sheet, but the only location for it to pull for base ability checks is the ability modifier cell which doesn't/shouldn't include these bonuses.

cogs5e/sheets/gsheet.py
thecoppersanctum: I assume then that this would necessitate changing the various instances of `self.version >= 2` to something else then as well?

mommothazaz123: Yup, Python handles this pretty easily:
```py
>>> (2, 1) >= (2, 0)
True
>>> (2, 0) >= (2, 1)
False
>>> (1, 0) >= (2, 0)
False
>>> (1, 4) >= (2, 0)
False
```

cogs5e/sheets/gsheet.py
mommothazaz123: For readability, can you pull base ability checks into its own block (might require changing some of the constants too) rather than doing it in a one-liner? Making some kind of loop dedicated to something like `BASE_ABILITY_CHECKS` makes sense to handle those special cases.

PR #1008, opened 2020-03-25
Add Great Weapon Fighting Support

Croebh:
### Summary
Automatically add the `ro<3` for the Great Weapon Fighting fighting style to two handed melee weapons.

#### PR Type

- [ ] This PR fixes an issue.
#### Other


Frogvall:
Even though I agree that the most sensible is to always reroll 1s and 2s with GWF, everyone might not agree. Some people *might* prefer to only automatically reroll 1s. Should there perhaps be a flag you can set for that?

PR #1245, opened 2020-09-13
Show changes when modifying CC values and add CC testing

### LINKED ISSUE #1202: AFR-601 Make `!cc +/-<#>` show the changes made
**Feature Request:** Make `!cc +/-<#>` show the changes made.
**Who would use it?:**  Everyone
**What part of the bot?:**  General Function - Custom Counters - `!cc`
**How it would work:**  `!cc "Lucky Dice" -1` Then will display the CC with a `-1` to indicate they went down 1 instead of just showing 2 (given they had 3 to start with)
**Why?:** For servers who have a lot of different players and systems. Custom counters can be used a lot without Aliases to display the changes to the counters.   Home brew classes that have counters or even new purchasing systems like Downtime Days.  when a user adjusts the amount using the `!cc +/-<#>` command it only displays the end results leaving staff members searching everywhere to make sure the amounts add up.

- MystedBlaze#0003

> Upvote - MystedBlaze#0003

> Upvote - silverbass#2407

> Upvote - Crys#4802

> Upvote - Swiftgale#4167

> Upvote - Firecrystal 101#0768

> Upvote - spamandtuna#9657

> Upvote - Lathaon#6649

> Upvote - Infinidoge#1337

> Upvote - Zekl#1505

> Upvote - eternalphoenix64#3855

> Upvote - Retr0#5626

> Upvote - Riptidecord#7191

> Upvote - Dr Turtle#1771

> Upvote - Overlord#9418

> Upvote - TheReverendB#1377

> Upvote - James Hurburgh#6265

> Upvote - Aether_Blackbolt#1361

> Upvote - Misfit.EVAD3#6839

Votes: +19 / -0

1drturtle:
### Summary
Shows changes to CC value with `!cc`
When using `!cc ... set` will show delta between old and new value
Resolves #1202 

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/gametrack.py
+++ b/cogs5e/gametrack.py
@@ -450,6 +450,7 @@ class GameTrack(commands.Cog):
             operator = m[0]
             modifier = m[-1]
 
+        change = ''
         try:
             modifier = int(modifier)
         except ValueError:
@@ -457,8 +458,11 @@ class GameTrack(commands.Cog):
         result_embed = EmbedWithCharacter(character)
         if not operator or operator == 'mod':
             new_value = counter.value + modifier
+            change = f' ({"+" if modifier > 0 else ""}{modifier})'
         elif operator == 'set':
             new_value = modifier
+            delta = modifier - counter.value
+            change = f' ({"+" if delta > 0 else ""}{delta})'
         else:
             return await ctx.send("Invalid operator. Use mod or set.")
 
@@ -470,7 +474,7 @@ class GameTrack(commands.Cog):
         else:
             out = str(counter)
 
-        result_embed.add_field(name=counter.name, value=out)
+        result_embed.add_field(name=f'{counter.name}{change}', value=out)
 
         await try_delete(ctx.message)
         await ctx.send(embed=result_embed)
--- a/cogs5e/gametrack.py
+++ b/cogs5e/gametrack.py
@@ -451,6 +451,7 @@ class GameTrack(commands.Cog):
             modifier = m[-1]
 
         change = ''
+        old_value = counter.value
         try:
             modifier = int(modifier)
         except ValueError:
@@ -458,11 +459,8 @@ class GameTrack(commands.Cog):
         result_embed = EmbedWithCharacter(character)
         if not operator or operator == 'mod':
             new_value = counter.value + modifier
-            change = f' ({"+" if modifier > 0 else ""}{modifier})'
         elif operator == 'set':
             new_value = modifier
-            delta = modifier - counter.value
-            change = f' ({"+" if delta > 0 else ""}{delta})'
         else:
             return await ctx.send("Invalid operator. Use mod or set.")
 
@@ -470,9 +468,22 @@ class GameTrack(commands.Cog):
         await character.commit(ctx)
 
         if new_value - counter.value:
-            out = f"{str(counter)}\n({abs(new_value - counter.value)} overflow)"
+            overflow = new_value - counter.value
+            changed = counter.value - old_value
+            out = f"{str(counter)}\n({abs(overflow)} overflow)"
+            # Calculations for Change (with overflow)
+            if not operator or operator == 'mod':
+                change = f' ({"+" if changed > 0 else ""}{changed})'
+            elif operator == 'set':
+                change = f' ({"+" if changed > 0 else ""}{changed})'
         else:
             out = str(counter)
+            # Calculations for Change (without overflow)
+            if not operator or operator == 'mod':
+                change = f' ({"+" if modifier > 0 else ""}{modifier})'
+            elif operator == 'set':
+                delta = modifier - old_value
+                change = f' ({"+" if delta > 0 else ""}{delta})'
 
         result_embed.add_field(name=f'{counter.name}{change}', value=out)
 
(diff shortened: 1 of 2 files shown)


Croebh:
How does this account for overflow?

1drturtle:
> How does this account for overflow?
Good point, will push a change for that here in a second

1drturtle:
@Croebh 
Added code that will handle the overflows, but the difference will always limit itself
For example, a CC with min 0 and max 10, if it was 5 and someone did `!cc set testing 15` it would be `(+5)`

cogs5e/gametrack.py
```diff
@@ -466,11 +468,24 @@ async def customcounter(self, ctx, name=None, *, modifier=None):
...
+                change = f' ({"+" if changed > 0 else ""}{changed})'
+            elif operator == 'set':
+                change = f' ({"+" if changed > 0 else ""}{changed})'
         else:
             out = str(counter)
-
-        result_embed.add_field(name=counter.name, value=out)
+            # Calculations for Change (without overflow)
+            if not operator or operator == 'mod':
+                change = f' ({"+" if modifier > 0 else ""}{modifier})'
```
Croebh, line 483: As a note, you can use `:+` in the f string to have it always show the sign:
`f' ({modifier:+})'` 

(Your way works, just thought I'd also point out an alternative you may not have been aware of)

1drturtle: I did not know that, thanks, will update!

1drturtle:
Quick Previews:
Discord_Alsc010npj
image

cogs5e/gametrack.py
```diff
@@ -466,11 +468,24 @@ async def customcounter(self, ctx, name=None, *, modifier=None):
         await character.commit(ctx)
 
         if new_value - counter.value:
-            out = f"{str(counter)}\n({abs(new_value - counter.value)} overflow)"
+            overflow = new_value - counter.value
+            changed = counter.value - old_value
+            out = f"{str(counter)}\n({abs(overflow)} overflow)"
+            # Calculations for Change (with overflow)
+            if not operator or operator == 'mod':
```
mommothazaz123, line 475: `operator` is guaranteed to exist here, so I don't think you need to do this check (tbf, my code is a little spaghetti. 😛)

Since one of the if bodies is guaranteed to be true, and they seem to have the same body, you might be able to take the delta code out and place it outside of the `if new_value - counter.value:`/`else` block?

cogs5e/gametrack.py
```diff
@@ -466,11 +468,24 @@ async def customcounter(self, ctx, name=None, *, modifier=None):
...
-
-        result_embed.add_field(name=counter.name, value=out)
+            # Calculations for Change (without overflow)
+            if not operator or operator == 'mod':
+                change = f' ({modifier:+})'
+            elif operator == 'set':
+                delta = modifier - old_value
+                change = f' ({delta:+})'
+
+        result_embed.add_field(name=f'{counter.name}{change}', value=out)
```
mommothazaz123, line 488: Personally, I'm more of a fan of having the delta display next to the value rather than in the title. (Consistent with `!g hp`)

1drturtle: `!g hp` Is just a text send instead of an embed, but I can definitely change that

cogs5e/gametrack.py
1drturtle: `operator` is None if the user doesn't specify `mod` or `set` (Ex `!cc testing -4`) So this check is for if it's not nothing or mod

cogs5e/gametrack.py
mommothazaz123: oh! well... you get the idea 😛

1drturtle: Like this?
image

mommothazaz123: yup!

1drturtle:
image

tests/cogs5e/gametrack_test.py
```diff
@@ -173,21 +174,50 @@ async def test_sb_remove(self, avrae, dhttp):
...
+        char = await active_character(avrae)
+        cc_embed = EmbedWithCharacter(char)
+        # Needed to allow for embed comparison
+        char.set_setting('color', 0xffffff)
+        cc_embed.color = 0xffffff
+
+        cc_embed.add_field(name=r'.+', value=r'((\d+)|(\d+\/\d+)) (\((\+|-)\d+\))(\n)*(\(\d+ .+\))*')
         avrae.message("!cc TESTCC +5")
+        await dhttp.receive_delete()
+        await dhttp.receive_message(embed=cc_embed)
```
mommothazaz123, line 204: it'd be cool to add an assertation on the state of the counter here and in other places!

```suggestion
        await dhttp.receive_message(embed=cc_embed)
        assert next(cc for cc in char.consumables if cc.name == "TESTCC").value == 5
```

mommothazaz123 (approved):

1drturtle:
Builds fails, don't know why. Code to prevent this fail seems to not be executing

mommothazaz123:
weird. try building the embed from scratch rather than using `EmbedWithCharacter`, only adding the relevant fields - the comparison code should ignore any fields that aren't present in the comparison embed

1drturtle:
> weird. try building the embed from scratch rather than using `EmbedWithCharacter`, only adding the relevant fields - the comparison code should ignore any fields that aren't present in the comparison embed

Alright, I'll try that. It's probably the source.

mommothazaz123 (approved):

PR #1249, opened 2020-09-16
Move Attack sort from str to initalization

1drturtle:
### Summary
Moves attack list sorting from `str` to class initialization. This fixes the issue of attacks being in different spots in the data vs the str.

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/models/sheet/attack.py
+++ b/cogs5e/models/sheet/attack.py
@@ -65,7 +65,7 @@ class AttackList:
     def __init__(self, attacks=None):
         if attacks is None:
             attacks = []
-        self.attacks = sorted((atk for atk in attacks), key=lambda atk: str(atk).lower())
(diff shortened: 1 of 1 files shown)


mommothazaz123:
I'm not sure this is the right direction to go, since this causes a semantic data change on character commit rather than just changing the display. Additionally, you'd have to do a resort on each modification operation (append, extend, etc).

Why not add the sorting code to `__str__()` as well instead?

1drturtle:
Alright, so this FR seems to have a lot of issues that I didn't think about, so I'm going to revert the sorting

mommothazaz123:
TL;DR:  After a conversation in Discord we determined that changing attack sorting on the `AttackList` level would cause issues in other `StatBlock` models like monsters.

PR #1285, opened 2020-10-02
Add descriptions to Init Effects

### LINKED ISSUE #890: AFR-465 The ability to set a description on `!init effect`s
**Feature Request:** The ability to set a description on `!init effect`s
**Who would use it?** Everyone from aliasers, homebrewers, dm's and users
**What part of the bot?** Initiative
**How would it work?** 
`!init effect OR1 Poison -dur 5 -desc "Description from !i effect go here!"`

`!init list`
> ```markdown
> Current initiative: 40 (round 1)
> ================================
> # 40: Thelen Sylvaranth <20/20 HP> (AC 12, Uber Spell [10 rds])
>    3: Orc
>      - OR1 <Healthy> (Poison [5 rds] <N>)
>      - OR2 <Healthy> (Poison [7 rds] <N>) 
>      - OR3 <Healthy> 
>      - OR4 <Healthy> 
>    2: Wolf
>      - WO1 <Healthy> 
>      - WO2 <Healthy> 
>      - WO3 <Healthy> (Poison [5 rds] <N>)
>      - WO4 <Healthy> 
>    1: GO1 <Healthy> 
>    1: GO2 <Healthy> 
> ```
`!init status OR1`
> ```markdown
> OR1 <Healthy>
> # Notes from !i note!
> * Poison [5 rounds]
>  - Description from !i effect go here!```
**Why should this be added?** Having the ability to tie a description to an effect is very useful, as you can include details about when you might save to end it, any effects that `!init effect` doesn't handle (like advantage on attacks, etc), and flavour text.

- Croebh#5603

> Upvote - Croebh#5603

> Upvote - TidusIV#2328

> Upvote - RiptideMatt#7191

> Upvote - Marcus#3244

> Upvote - Infinidoge#1337

> Upvote - Zekl#9787

> Upvote - Lukas#0386

> Upvote - silverbass#2407

> Upvote - AriAlucard#4612

> Upvote - li8ning#8533

Votes: +11 / -0

Croebh:
### Summary
This adds a `-desc` argument to `!init effect`, as well as to automation and `SimpleCombatant.add_effect()`.

I still need to add this to the actual Automation editor for spells and attacks, but I confirmed it works with `!attack import` using a modified JSON to add the `desc` tag to an ieffect.

image

image

Resolves #890 (AFR-465)

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/initTracker.py
+++ b/cogs5e/initTracker.py
@@ -925,7 +925,8 @@ class InitTracker(commands.Cog):
         -neutral <damage type> - Removes the combatant's immunity, resistance, or vulnerability to the given damage type.
         __General__
         -ac <ac> - modifies ac temporarily; adds if starts with +/- or sets otherwise.
-        -sb <save bonus> - Adds a bonus to all saving throws."""
+        -sb <save bonus> - Adds a bonus to all saving throws.
+        -desc <"description"> - Adds a description of the effect."""
         combat = await Combat.from_ctx(ctx)
         args = argparse(args)
 
@@ -942,6 +943,7 @@ class InitTracker(commands.Cog):
         conc = args.last('conc', False, bool)
         end = args.last('end', False, bool)
         parent = args.last('parent')
+        desc = args.last('desc')
 
         if parent is not None:
             parent = parent.split('|', 1)
@@ -957,7 +959,7 @@ class InitTracker(commands.Cog):
                 out = "Effect already exists."
             else:
                 effect_obj = Effect.new(combat, combatant, duration=duration, name=effect_name, effect_args=args,
-                                        concentration=conc, tick_on_end=end)
+                                        concentration=conc, tick_on_end=end, desc=desc)
                 result = combatant.add_effect(effect_obj)
                 if parent:
                     effect_obj.set_parent(parent)
--- a/cogs5e/initTracker.py
+++ b/cogs5e/initTracker.py
@@ -926,7 +926,7 @@ class InitTracker(commands.Cog):
         __General__
         -ac <ac> - modifies ac temporarily; adds if starts with +/- or sets otherwise.
         -sb <save bonus> - Adds a bonus to all saving throws.
-        -desc <"description"> - Adds a description of the effect."""
+        -desc <description> - Adds a description of the effect."""
         combat = await Combat.from_ctx(ctx)
         args = argparse(args)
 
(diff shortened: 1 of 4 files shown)


1drturtle (approved):

Croebh:
I just also added a character limit (500, though the number was chosen arbitrarily), and annotated it.

cogs5e/initTracker.py
```diff
@@ -925,7 +925,8 @@ async def effect(self, ctx, target_name: str, effect_name: str, *args):
         -neutral <damage type> - Removes the combatant's immunity, resistance, or vulnerability to the given damage type.
         __General__
         -ac <ac> - modifies ac temporarily; adds if starts with +/- or sets otherwise.
-        -sb <save bonus> - Adds a bonus to all saving throws."""
+        -sb <save bonus> - Adds a bonus to all saving throws.
+        -desc <"description"> - Adds a description of the effect."""
```
mommothazaz123, line 929: why the quotes here?

cogs5e/models/automation.py
```diff
@@ -947,7 +954,7 @@ def run(self, autoctx):
                 autoctx.queue(f"**Concentration**: dropped {', '.join([e.name for e in conc_conflict])}")
         else:
             effect = init.Effect.new(None, None, self.name, duration, autoctx.parse_annostr(self.effects),
-                                     tick_on_end=self.tick_on_end, concentration=self.concentration)
+                                     tick_on_end=self.tick_on_end, concentration=self.concentration, desc=self.desc)
```
mommothazaz123, line 957: should this be `desc` rather than `self.desc` to get the rendered description?

cogs5e/initTracker.py
Croebh: To represent that you need quotes if its more than one word. Followed the format from the attack arg in `!init effect`

mommothazaz123: I guess. Feels a little weird, though, since usually the requirement for quotes in an argument is implicit (see !c -phrase for example - they're only there in the attack arg because it has a format that needs to be followed - honestly I'm not 100% on that docs format either, but eh)

cogs5e/models/automation.py
Croebh: Yes, yes it should be

mommothazaz123 (approved):

PR #1322, opened 2020-10-17
Grab the 'Additional' sheet directly, instead of using index

Croebh:
### Summary
This changes the gsheet importer to grab the sheet named `Additional` instead of using an index. Should reduce the amount of errors people get due to accidentally adding sheets or reordering them.

#### PR Type

- [ ] This PR fixes an issue.
#### Other


mommothazaz123 (approved):

mommothazaz123:
Out of curiosity, is there a way to validate this gsheet-side? E.g. to display an error if the sheet is renamed

cc @thecoppersanctum

Infinidoge:
I don't believe there is a built-in function to get the sheet name, however it should be possible through AppsScript and the ScriptEditor built into gsheet

```js
function sheetName() {
  var ss = SpreadsheetApp.getActiveSpreadsheet();
  var sheet = ss.getActiveSheet();
  return sheet.getName();
}
```
(found on a google support thread [here](https://support.google.com/docs/thread/8603869?hl=en&msgid=8768429))

which can then be used in a cell as per normal functions

Not sure if there is any objections to including script elements into the gsheet.

PR #1329, opened 2020-10-20
Add description and title to consumables

### LINKED ISSUE #1276: AFR-639 Addition to `!cc create`: `-title` and `-desc`
**Feature Request:** Addition to `!cc create`: `-title` and `-desc`
**Who would use it?** Everyone
**What part of the bot?** Custom counters (so character model?)
**How would it work?** Specifying `-title "{name} does a Thing!"` and `-desc "Mechanics Text"` when running `!cc create`, and then having the title and desc display when using `!cc <counter> -1`
**Why should this be added?** Currently, you need to write an alias to display any text with cc manipulation. Having this just be built in to `!cc` will make it easier for users to set up basic non-automated things.

For example, Divine Sense and War Priest don't have any automation, and just marking off a use and displaying mechanical text would handle them quite well.

- silverbass#2407

> Upvote - Dr Turtle#1771

> Upvote - spamandtuna#9657

> Upvote - Lathaon#6649

> Upvote - Spooky Riptidecord#7191

> Upvote - Hopop#6751

> Upvote - Derixyleth#0636

> Upvote - Zekl#1505

> Upvote - Robbe#7964

> Upvote - Roedelius#9141

> Upvote - Mobius#1442

> Upvote - Ithala#3408

> Upvote - syrup#1965

Votes: +12 / -0

1drturtle:
Summary
======
Adds a description and title to custom counters/consumables
In the title, `[name]` is replaced with the character name and `[cname]` is replace with the custom counter name.

Resolves #1276 (AFR-639)

Methods added
----
`AliasCustomCounter` -
* `AliasCustomCounter.title`
* `AliasCustomCounter.desc`

`CustomCounter` -
* `CustomCounter.title`
* `CustomCounter.desc`

Checklist
======

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/gametrack.py
+++ b/cogs5e/gametrack.py
@@ -446,13 +446,11 @@ class GameTrack(commands.Cog):
         # replace [name] in title
         cc_embed_title = cc_embed_title.replace('[name]', character.name)
 
-        cc_desc = counter.desc if counter.desc else ''
-
         if modifier is None:  # display value
             counter_display_embed = EmbedWithCharacter(character)
             counter_display_embed.add_field(name=counter.name, value=counter.full_str())
-            if cc_desc:
-                counter_display_embed.add_field(name='Description', value=cc_desc, inline=False)
+            if counter.desc:
+                counter_display_embed.add_field(name='Description', value=counter.desc, inline=False)
             return await ctx.send(embed=counter_display_embed)
 
         operator = None
@@ -486,8 +484,8 @@ class GameTrack(commands.Cog):
 
         result_embed.add_field(name=cc_embed_title, value=out)
 
-        if cc_desc:
-            result_embed.add_field(name='Description', value=cc_desc, inline=False)
+        if counter.desc:
+            result_embed.add_field(name='Description', value=counter.desc, inline=False)
 
         await try_delete(ctx.message)
         await ctx.send(embed=result_embed)
--- a/cogs5e/gametrack.py
+++ b/cogs5e/gametrack.py
@@ -441,12 +441,16 @@ class GameTrack(commands.Cog):
         character: Character = await Character.from_ctx(ctx)
         counter = await character.select_consumable(ctx, name)
 
-        cc_embed_title = counter.title if counter.title else counter.name
+        cc_embed_title = counter.title if counter.title is not None else counter.name
+
+        # replace [cname] and [name] in title
+        cc_embed_title = cc_embed_title.replace('[name]', character.name).replace('[cname]', counter.name)
+
         cc_desc = counter.desc + '\n' if counter.desc else ''
 
         if modifier is None:  # display value
             counter_display_embed = EmbedWithCharacter(character)
-            counter_display_embed.add_field(name=cc_embed_title, value=counter.full_str(desc=True))
+            counter_display_embed.add_field(name=counter.name, value=counter.full_str(desc=True))
             return await ctx.send(embed=counter_display_embed)
 
         operator = None
@@ -478,7 +482,7 @@ class GameTrack(commands.Cog):
         else:
             out = f"{str(counter)} {delta}"
 
-        result_embed.add_field(name=counter.name, value=cc_desc + out)
+        result_embed.add_field(name=cc_embed_title, value=cc_desc + out)
 
         await try_delete(ctx.message)
         await ctx.send(embed=result_embed)
--- a/cogs5e/gametrack.py
+++ b/cogs5e/gametrack.py
@@ -443,8 +443,8 @@ class GameTrack(commands.Cog):
 
         cc_embed_title = counter.title if counter.title is not None else counter.name
 
-        # replace [cname] and [name] in title
-        cc_embed_title = cc_embed_title.replace('[name]', character.name).replace('[cname]', counter.name)
+        # replace [name] in title
+        cc_embed_title = cc_embed_title.replace('[name]', character.name)
 
         cc_desc = counter.desc + '\n' if counter.desc else ''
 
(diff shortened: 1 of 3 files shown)


cogs5e/models/sheet/player.py
```diff
@@ -94,9 +94,13 @@ class CustomCounter:
                  'hp': "Gaining HP"}
 
     def __init__(self, character, name, value, minv=None, maxv=None, reset=None, display_type=None, live_id=None,
-                 reset_to=None, reset_by=None):
+                 title=None, desc=None, reset_to=None, reset_by=None):
```
mommothazaz123, line 97:
```suggestion
                 reset_to=None, reset_by=None, title=None, desc=None):
```

I don't think `__init__` is ever called with only positional args, but this prevents breaking it if it is.

cogs5e/models/sheet/player.py
```diff
@@ -116,12 +120,12 @@ def from_dict(cls, char, d):
...
         return {"name": self.name, "value": self._value, "minv": self.min, "maxv": self.max, "reset": self.reset_on,
-                "display_type": self.display_type, "live_id": self.live_id, "reset_to": self.reset_to,
-                "reset_by": self.reset_by}
+                "display_type": self.display_type, "live_id": self.live_id, "title": self.title, "desc": self.desc,
+                "reset_to": self.reset_to, "reset_by": self.reset_by}
 
     @classmethod
     def new(cls, character, name, minv=None, maxv=None, reset=None, display_type=None, live_id=None,
-            reset_to=None, reset_by=None):
+            title=None, desc=None, reset_to=None, reset_by=None):
```
mommothazaz123, line 128:
```suggestion
            reset_to=None, reset_by=None, title=None, desc=None):
```

cogs5e/models/sheet/player.py
```diff
@@ -168,7 +172,18 @@ def new(cls, character, name, minv=None, maxv=None, reset=None, display_type=Non
...
+            raise InvalidArgument('Description must be less than 800 characters.')
+
+        if title and len(title) >= 256:
+            raise InvalidArgument('Title must be less than 256 characters.')
+
+        if len(name) > 256:
+            raise InvalidArgument('Name must be less than 256 characters.')
+
+        return cls(character, name.strip(), initial_value, minv, maxv, reset, display_type, live_id,
+                   title, desc, reset_to, reset_by)
```
mommothazaz123, line 186:
```suggestion
                   reset_to, reset_by, title, desc)
```
to match the positional arg suggestions

cogs5e/models/sheet/player.py
```diff
@@ -196,6 +211,25 @@ def get_reset_to(self):
     def value(self):
         return self._value
 
+    @property
```
mommothazaz123, line 214: I'm not entirely sure these properties (and similarly, `_title` and `_desc` being private attrs) are entirely necessary.

As is, `CustomCounter.title` (the property here) is what is called when it's committed to the database, so the `[cname]` and `[name]`s are lost - I don't believe `[cname]` and `[name]` are necessary, though.

cogs5e/models/sheet/player.py
```diff
@@ -274,14 +312,15 @@ def full_str(self):
     def __str__(self):
         _max = self.get_max()
 
+        out = self.desc + '\n' if self.desc else ''
```
mommothazaz123, line 315: This'll put the description (up to 800 characters) in *every* location the counter is stringified. That doesn't seem necessary - for example, if I had an alias 
```
!alias wildmagic embed <drac2>
out = [...]
# some wild magic code here
out.append(f'-f "Some Resource|{{character().cc_str('Some Resource')}}"')
return '\n'.join(out)
```

I wouldn't want the description to appear.

mommothazaz123 (changes requested):
Thanks for the contribution! I'm not entirely sure what the scope of the feature request this PR is closing is. Should titles and descriptions appear everywhere? In rests? Only when they're manually changed with `!cc`?

I believe the original scope of the FR only encompasses the latter (and thus that the title and desc handling should be in that command rather than in methods on the cc object), and that the `[name]` and `[cname]` modification is unnecessary.

1drturtle:
Updated - 
Removed description from `__str__`
Fixed `title` and `desc` properties (`[name]` and `[cname]` are still supported, but are replaced in the command instead.)
Moved order of `title` and `desc`
image
(In aliasing)
image

RFC - Remove `[cname]` since when it's created they know the name of it?

mommothazaz123:
Screenshot 2020-10-26 at 11 43 51

Maybe the description should have its own field (under the meta of cc info/cc value) - this looks a little cluttered

1drturtle:
With or without description header? I'm leaning more towards with header so I that's what I committed.
image

cogs5e/models/sheet/player.py
```diff
@@ -252,11 +267,13 @@ def full_str(self):
...
 
+        val = ''
+
         if self.display_type == 'bubble':
             assert self.max is not None
-            val = f"{bubble_format(self.value, _max)}\n"
+            val += f"{bubble_format(self.value, _max)}\n"
         else:
-            val = f"**Current Value**: {self.value}\n"
+            val += f"**Current Value**: {self.value}\n"
```
mommothazaz123, line 276: This change can probably be reverted

cogs5e/gametrack.py
```diff
@@ -434,15 +434,25 @@ async def spellbook_remove(self, ctx, *, spell_name):
...
             return await self.customcounter_summary(ctx)
         character: Character = await Character.from_ctx(ctx)
         counter = await character.select_consumable(ctx, name)
 
+        cc_embed_title = counter.title if counter.title is not None else counter.name
+
+        # replace [name] in title
+        cc_embed_title = cc_embed_title.replace('[name]', character.name)
+
+        cc_desc = counter.desc if counter.desc else ''
```
mommothazaz123, line 449: The use of `cc_desc` below can probably be changed to just checks on `counter.desc`, and this variable removed

mommothazaz123 (approved):

PR #1336, opened 2020-10-23
Add channel category to AliasContext

### LINKED ISSUE #1333: AFR-668 `ctx.category.id` to return discord category id
**Feature Request:** `ctx.category.id` to return discord category id
**Who would use it?** Aliasers
**What part of the bot?** Aliasing
**How would it work?** Simply gets the category of the channel the alias is run in
**Why should this be added?** To check if aliases are run in a group of channels that could be created in the future

- alpha#1832

> Upvote - Infinidoge#1337

> Upvote - alpha#1832

> Upvote - Mobius#1442

> Upvote - Zekl#1505

> Upvote - Agathist#1706

> Upvote - Mahkasad#5996

Votes: +6 / -0

Infinidoge:
# Summary
Adds a property to AliasContext to represent the category of the channel that it has been executed in.

Resolves #1333 

## Methods added

`AliasChannel` -

* `AliasChannel.category`

`AliasCategory` -

* `AliasCategory.name`
* `AliasCategory.id`

# Checklist

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/aliasing/api/context.py
+++ b/aliasing/api/context.py
@@ -71,8 +71,8 @@ class AliasContext:
 
 
     def __repr__(self):
-        return f"<AliasContext guild={self.guild} channel={self.channel} category={self.category} author={self.author}"\
-               f" prefix={self.prefix} alias={self.alias}>"
+        return f"<AliasContext guild={self.guild} channel={self.channel} author={self.author} prefix={self.prefix} " \
+               f"alias={self.alias}>"
 
 
 class AliasGuild:
@@ -121,8 +121,7 @@ class AliasChannel:
         self._name = str(channel)
         self._id = channel.id
         self._topic = channel.topic if not isinstance(channel, discord.DMChannel) else None
-        self._category = AliasCategory(channel.category) if hasattr(channel, 'category') \
-                                                            and channel.category is not None else None
+        self._category = AliasCategory(channel.category) if getattr(channel, 'category', None) is not None else None
 
     @property
     def name(self):
@@ -233,7 +232,7 @@ class AliasCategory:
     @property
     def name(self):
         """
-        The name of the category, not including the preceding hash (#).
+        The name of the category
 
         :rtype: str
         """
--- a/aliasing/api/context.py
+++ b/aliasing/api/context.py
@@ -15,8 +15,6 @@ class AliasContext:
         """
         self._guild = None if ctx.guild is None else AliasGuild(ctx.guild)
         self._channel = AliasChannel(ctx.channel)
-        self._category = AliasCategory(ctx.channel.category) if hasattr(ctx.channel, 'category') \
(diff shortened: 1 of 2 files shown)


1drturtle:
Maybe include this under `ctx.channel`?

Infinidoge:
That would make sense with the Discord.py model it comes from, 

I do however think it would be better as it's own attribute of `AliasContext` to reduce semi-unecessary nesting.
I also think `ctx.category` is more intuitive to the average user than `ctx.channel.category`

mommothazaz123:
I agree that it should be under `AliasChannel` rather than `AliasContext` at the top level.

Infinidoge:
Moved, updated PR description to match.
Currently updating the documentation

aliasing/api/context.py
```diff
@@ -69,9 +69,10 @@ def alias(self):
         """
         return self._alias
 
+
     def __repr__(self):
-        return f"<AliasContext guild={self.guild} channel={self.channel} author={self.author} prefix={self.prefix}" \
-               f" alias={self.alias}>"
+        return f"<AliasContext guild={self.guild} channel={self.channel} category={self.category} author={self.author}"\
```
mommothazaz123, line 74: `self.category` is not defined here

aliasing/api/context.py
```diff
@@ -120,6 +121,8 @@ def __init__(self, channel):
         self._name = str(channel)
         self._id = channel.id
         self._topic = channel.topic if not isinstance(channel, discord.DMChannel) else None
+        self._category = AliasCategory(channel.category) if hasattr(channel, 'category') \
+                                                            and channel.category is not None else None
```
mommothazaz123, line 125: could be simplified to `getattr(channel, 'category', None)`

aliasing/api/context.py
```diff
@@ -204,3 +216,37 @@ def display_name(self):
...
+        """
+        :type channel: discord.ChannelCategory
+        """
+        self._name = str(category)
+        self._id = category.id
+
+    @property
+    def name(self):
+        """
+        The name of the category, not including the preceding hash (#).
```
mommothazaz123, line 236: what preceding hash?

mommothazaz123 (changes requested):

aliasing/api/context.py
Infinidoge: Just something I missed when editing the documentation 
(the functions were initially copied from the channels class)

Fixing in next commit.

aliasing/api/context.py
Infinidoge: TIL `getattr` has a default section

```py
self._category = AliasCategory(channel.category) if getattr(channel, 'category', None) is not None else None
```

Would be the proper way to apply this, correct?

mommothazaz123 (approved):

PR #1371, opened 2020-11-22
Update evaluators.py

### LINKED ISSUE #1370: AVR-681 get_svar() returns NoneType in private messages
**Describe the bug**
calling `get_svar()` in a private message with Avrae returns NoneType, even with a default provided

**Severity**
Medium

**Context**
Any alias that uses `get_svar()`

**To Reproduce**
1. use `get_svar()` in scripting
2. Provide a default argument
3. run the command in private message with Avrae

**Expected behavior**
The default argument should be returned instead of NoneType

rtotheichie:
Resolves #1370 

### Summary
Returns a default argument if one is provided when get_svar is used in private messages

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/aliasing/evaluators.py
+++ b/aliasing/evaluators.py
@@ -342,7 +342,7 @@ class ScriptingEvaluator(draconic.DraconicInterpreter):
         :rtype: str or None
         """
         if self.ctx.guild is None:
-            return None
+            return default
         if name not in self._cache['svars']:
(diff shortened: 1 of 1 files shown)


1drturtle:
I believe the reason it returned None when there was no guild was to differentiate between not having the svar and being in a DM, but since we can do that with the context methods I think this is fine.

mommothazaz123 (approved):

PR #1383, opened 2020-11-27
Allow aliases to set combatant groups

### LINKED ISSUE #1317: AFR-660 Add a set_group() function to Aliasing API
Feature Request: Add a set_group() function to Aliasing API
Who would use it? anyone who may be trying to create interesting new spells that use grouping in combat. Charming a monster, creating illusions, shuffling your character and illusions with your name in the group
What part of the bot? Alias
How would it work? just like set_name() 
Why should this be added? I have been working on creating a few spell implementations, in this case it is very relevant for mirror image. The way it is working now, im using a multiline of `!init opt name1 -name name2 -group groupname !init opt name2 -name name3 -group none` to shuffle a character in a group with 3 other illusions. Because of the multiline delay the spell is  looking clunky and easy to cheat. Also for consistency since there is a set_name() and a bunch of other commands that allow us to control a SimpleCombatant settings.

- fenyv#1430

> Upvote - Croebh#5603

> Upvote - Lathaon#6649

> Upvote - Mobius#1442

> Upvote - Robbe#7964

> Upvote - Crys#4802

> Upvote - Brinjal#4037

Votes: +7 / -0

1drturtle:
Summary
======
Allows aliases to set the group of a combatant.

Adds `Combatant.set_group(group_name: str)`
Changes `!i opt` to use the new method on Combatant
Adds `SimpleCombatant.set_group(group: str)` which uses `Combatant.set_group`

Issues
------

Resolves #1317 

Checklist
=====

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/aliasing/api/combat.py
+++ b/aliasing/api/combat.py
@@ -236,26 +236,7 @@ class SimpleCombatant(AliasStatBlock):
         if not isinstance(group, str):
             raise ValueError('Group name must be a string.')
 
-        # copied and modified from !i opt code
-
-        combat = self._combatant.combat
-        combatant = self._combatant
-        current = combatant.combat.current_combatant
-        was_current = combatant is current or \
-                      (isinstance(current, CombatantGroup) and combatant in current and len(current) == 1)
-
-        combat.remove_combatant(combatant, ignore_remove_hook=True)
-        if group.lower() == 'none':
-            combat.add_combatant(combatant)
-            if was_current:
-                combat.goto_turn(combatant, True)
-            return None
-        else:
-            c_group = combat.get_group(group, create=combatant.init)
-            c_group.add_combatant(combatant)
-            if was_current:
-                combat.goto_turn(combatant, True)
-            return SimpleGroup(c_group)
+        return self._combatant.set_group(group_name=group)
 
     def set_note(self, note: str):
         """
--- a/aliasing/api/combat.py
+++ b/aliasing/api/combat.py
@@ -225,6 +225,38 @@ class SimpleCombatant(AliasStatBlock):
             raise ValueError("Combatants must have a name.")
         self._combatant.name = str(name)
 
+    def set_group(self, group: str):
+        """
+        Sets the combatant's group
+
+        :param str group: The name of the group. "None" to remove from group.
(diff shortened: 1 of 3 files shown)


mommothazaz123:
In general I'm not really comfortable with aliases modifying the combat state directly like this - I think at least the group changing code should be made an instance method on the `Combatant` object.

1drturtle:
> In general I'm not really comfortable with aliases modifying the combat state directly like this

That makes sense, but adding this would help with reducing the amount of multilines that are needed. Maybe we could change it to pass a `SimpleGroup`, so that the group would have to already exist, but this kind of defeats the purpose.

> I think at least the group changing code should be made an instance method on the Combatant object.

can do

cogs5e/models/initiative.py
```diff
@@ -688,6 +688,23 @@ def group(self):
     def group(self, value):
         self._group = value
 
+    def set_group(self, group_name: str):
+        current = self.combat.current_combatant
+        was_current = self is current or \
+                      (isinstance(current, CombatantGroup) and self in current and len(current) == 1)
+        self.combat.remove_combatant(self, ignore_remove_hook=True)
+        if group_name.lower() == 'none':
```
mommothazaz123, line 696: While we're here, may as well make this accept a `None` group name to clear groups rather than the string `'none'`

mommothazaz123 (approved):

aliasing/api/combat.py
```diff
@@ -225,6 +225,19 @@ def set_name(self, name: str):
...
+        Sets the combatant's group
+
+        :param str group: The name of the group. None to remove from group.
+        :return: The combatant's new group, or None if the combatant was removed from a group.
+        :rtype: :class:`~aliasing.api.combat.SimpleGroup` or None   
+        """
+        if not (isinstance(group, str) or group is None):
+            raise ValueError('Group name must be a string or None.')
+
+        return self._combatant.set_group(group_name=group)
```
mommothazaz123, line 239: This will return a normal `Group` rather than `SimpleGroup` if the combatant is added to a group

mommothazaz123 (approved):

PR #1427, opened 2021-01-14
Fix AVR-702 (server collection commands erroring in DMs)

### LINKED ISSUE #1415: AVR-702 `!serveralias subscribe` has unhelpful error message if you run it in avrae private messages
**What is the bug?**: `!serveralias subscribe` has unhelpful error message if you run it in avrae private messages
**Severity**: Low
**Steps to reproduce**: do `!serveralias subscribe any_workshop_url` in avrae private messages
**Context**: `Error: Command raised an exception: AttributeError: 'User' object has no attribute 'guild_permissions'` (i know why it does not work, it's just that the message could be more helpful.)

- mrad#9260
Verification: 0

1drturtle:
Summary
======

Adds checks to `_servalias_before_edit` and `_servsnippet_before_edit` to prevent them from executing in DM Channels.

Fixes #1415 (AVR-702)

Checklist
======

PR Type
---------

- [X] This PR fixes an issue.

Other
------

### DIFF
--- a/cogsmisc/customization.py
+++ b/cogsmisc/customization.py
@@ -309,6 +309,8 @@ async def _alias_before_edit(ctx, name=None):
 
 
 async def _servalias_before_edit(ctx, name=None):
+    if ctx.guild is None:
+        raise commands.NoPrivateMessage()
     if not _can_edit_servaliases(ctx):
         raise NotAllowed("You do not have permission to edit server aliases. Either __Administrator__ "
                          "Discord permissions or a role named \"Server Aliaser\" or \"Dragonspeaker\" "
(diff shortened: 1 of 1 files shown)


mommothazaz123:
Weird. This shouldn't be necessary - the definition of the `servalias` and `servsnippet` command groups already include a `guild_only_check` (see L440); I'd prefer investigating to see why that might not be running if possible, rather than adding a theoretically redundant check.

1drturtle:
I've looked through the code and I can't tell why this bug is occurring. It's a command that should have the checks. The only thing I can think of is the feature-flag decorator messing up the checks on the commands.

mommothazaz123:
Fixed in ce1c44b - as it turns out adding a check via decorator completely overwrites checks added via kwargs instead of adding the two together. (See https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/core.py#L247)

PR #1483, opened 2021-03-29
Fix automatic hit on crit arg

1drturtle:
Summary
======

Previously, the `crit` arg would auto-hit the target, even on a natural one (crit fail)
This change fixes that bug

Resolves AVR-764 (Not in GitHub yet)

Checklist
=======

PR Type
--------

- [X] This PR fixes an issue.
Other
-------

### DIFF
--- a/cogs5e/models/automation/effects.py
+++ b/cogs5e/models/automation/effects.py
@@ -317,14 +317,14 @@ class Attack(Effect):
                 ac = ac or autoctx.target.ac
 
             # assign hit values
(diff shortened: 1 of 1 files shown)


zhudotexe:
This doesn't appear to actually fix the linked issue (e.g. if the attack roll misses normally, without critically failing, the crit arg will still cause an automatic hit).

I've implemented a fix in 80e510d1fc1c16c2e745f875a65be202bb92e22e.

PR #1499, opened 2021-04-14
adding <, > and == for the aliasskill

### LINKED ISSUE #1500: AFR-787 Add support for the comparison operators (`>`, `<`, `==`, etc) to `AliasSkill`,  to compare with it's `AliasSkill.value`
**Feature Request:** Add support for the comparison operators (`>`, `<`, `==`, etc) to `AliasSkill`,  to compare with it's `AliasSkill.value`
**Who would use it?** Comparing the values of a skill while still keeping the skill object intact, such as with `min` and `max`
**What part of the bot?** Aliasing
**How would it work?** The comparison operators would compare the `AliasSkill.value` to `other`
**Why should this be added?** If you're making something that has a choice of skills and want to take the higher of the 2 options, a pythonic way would be to use `max(AliasSkill, AliasSkill)`, however this does not work due to the lack of comparison support. Furthermore, the addition of comparison support would allow for sorting lists including an `AliasSkill` object

- Infinidoge#1337

> Upvote - Dr Turtle#1771

> Upvote - spamandtuna#9657

> Upvote - Riptidecord#7191

> Upvote - Infinidoge#1337

> Upvote - Mobius#1442

> Upvote - merien#7290

Votes: +8 / -0

driesceuppens:
### Summary
Resolves #1500 
Adding support for the <, > and == operators in the aliasskill class.

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/aliasing/api/statblock.py
+++ b/aliasing/api/statblock.py
@@ -477,16 +477,13 @@ class AliasSkill:
                f"adv={self.adv!r}>"
     
     def __gt__(self, other):
-        if isinstance(other, AliasSkill):
-            return self.value > other.value
+        return isintance(other, AliasSkill) and self.value > other.value
     
-    def __lt__(self, other):
-        if isinstance(other, AliasSkill):
-            return self.value < other.value
-    
+    def __ge__(self, other):
+        return isintance(other, AliasSkill) and self.value >= other.value
+
     def __eq__(self, other):
-        if isinstance(other, AliasSkill):
-            return self.value == other.value
+        return isintance(other, AliasSkill) and self.value == other.value
 
 class AliasSkills:
     """
--- a/aliasing/api/statblock.py
+++ b/aliasing/api/statblock.py
@@ -475,7 +475,18 @@ class AliasSkill:
     def __repr__(self):
         return f"<{self.__class__.__name__} value={self.value!r} prof={self.prof!r} bonus={self.bonus!r} " \
                f"adv={self.adv!r}>"
-
+    
+    def __gt__(self, other):
+        if isinstance(other, AliasSkill):
+            return self.value > other.value
+    
+    def __lt__(self, other):
+        if isinstance(other, AliasSkill):
+            return self.value < other.value
+    
+    def __eq__(self, other):
+        if isinstance(other, AliasSkill):
+            return self.value == other.value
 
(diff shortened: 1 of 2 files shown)


driesceuppens:
I just noticed Dr. Turtle's comment, Might as well add those. Sorry for being clumsy 😄

zhudotexe:
Also technically, the presence of `__gt__` is not necessary (Python implicitly handles when the operation is only defined for one operand):
```py
class A:
    def __init__(self, v):
        self.v = v
    def __lt__(self, other):
        return isinstance(other, A) and self.v < other.v
    def __eq__(self, other):
        return isinstance(other, A) and self.v == other.v
    
one = A(1)
two = A(2)
one < two
# True
one > two
# False
one == two
# False
one <= two
# Traceback (most recent call last):
#   File "<input>", line 1, in <module>
# TypeError: '<=' not supported between instances of 'A' and 'A'
```

(Also, you can add more commits to the branch instead of closing and reopening the PR)

Infinidoge:
If you define the greater than, it no longer needs the `isinstance` check though, since it acts as the inverse operator.

```py
class A:
    def __init__(self, value):
        self.value = value

    def __lt__(self, other):
        return self.value < other

    def __gt__(self, other):
        return self.value > other


class B:
    def __init__(self, value):
        self.value = value

    def __lt__(self, other):
        return self.value < other

A1 = A(1)
A2 = A(2)

B1 = B(1)
B2 = B(2)

try:
    assert A1 < A2
except TypeError:
    print("A1 < A2: doesn't work")
else:
    print("A1 < A2: works")

try:
    assert B1 < B2
except TypeError:
    print("B1 < B2: doesn't work")
else:
    print("B1 < B2: works")
```
Results in:
```
A1 < A2: works
B1 < B2: doesn't work
```

Infinidoge:
(Additionally, this allows for proper comparison with other types that can compare ints)

driesceuppens:
So what would be better, adding all 5 or having the isinstance checks?

zhudotexe:
Doge just showed me https://rszalski.github.io/magicmethods/#comparisons - it looks like maybe `__cmp__` is the best choice

zhudotexe:
I lied, it was removed in Python 3

PR #1506, opened 2021-04-21
Add-Dice-Rolling-to-cc-modify

storytellermahkasad:
### Summary

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/gametrack.py
+++ b/cogs5e/gametrack.py
@@ -427,8 +427,18 @@ class GameTrack(commands.Cog):
     @commands.group(invoke_without_command=True, name='customcounter', aliases=['cc'])
     async def customcounter(self, ctx, name=None, *, modifier=None):
         """Commands to implement custom counters.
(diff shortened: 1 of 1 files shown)


cogs5e/gametrack.py
```diff
@@ -471,10 +486,9 @@ async def customcounter(self, ctx, name=None, *, modifier=None):
         await character.commit(ctx)
 
         delta = f"({counter.value - old_value:+})"
+        out = f"{str(counter)} {delta}\nRoll: {roll_result}"
```
zhudotexe, line 489: Will this display the roll even if the modifier is not a roll (e.g. `!cc blah -1`)?

storytellermahkasad: Currently yes, I considered making it skip the roll portion if it was just an int instead. Would that be preferable?

zhudotexe (approved):

PR #1527, opened 2021-06-06
Adds rollstring for -n, uses existing errors

storytellermahkasad:
### Summary
Adds roll string handling for `!init madd -n`
Resolves AFR-815

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/initTracker.py
+++ b/cogs5e/initTracker.py
@@ -233,7 +233,7 @@ class InitTracker(commands.Cog):
         init_skill = monster.skills.initiative
 
         combat = await Combat.from_ctx(ctx)
-        
+
         out = ''
         to_pm = ''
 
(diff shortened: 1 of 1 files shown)


zhudotexe (approved):
LGTM - it might be worth creating a util `int_or_dice` that can be plugged into the `type_` arg of `args.last`

Also, out of curiosity, how does this handle invalid dice strings/negatives?

storytellermahkasad:
> LGTM - it might be worth creating a util `int_or_dice` that can be plugged into the `type_` arg of `args.last`
> 
> Also, out of curiosity, how does this handle invalid dice strings/negatives?

That would be very nice to have, and invalid dice strings are handled with the same error as everywhere else. `Error in roll: Unexpected input on line 1, col 1: expected DECIMAL, U_OP, INTEGER, LPAR, D, got t`

Negatives are caught by: `recursion = 25 if n > 25 else 1 if n < 1 else n` which hasn't changed.

PR #1528, opened 2021-06-06
Add slash command check

storytellermahkasad:
### Summary
Adds Discord Slash-command conflict checking to prevent Avrae from becoming non-responsive to a prefix.

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogsmisc/customization.py
+++ b/cogsmisc/customization.py
@@ -359,7 +359,7 @@ class Customization(commands.Cog):
     async def prefix(self, ctx, prefix: str = None):
         """Sets the bot's prefix for this server.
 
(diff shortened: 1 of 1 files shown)


zhudotexe (commented):
imo, this should be a warning + confirm rather than a hard restriction for if some user wants to do it anyway

storytellermahkasad:
> imo, this should be a warning + confirm rather than a hard restriction for if some user wants to do it anyway

My testing of it, though brief, showed Avrae stopped responding entirely as soon as I had any prefix that began with a slash.

zhudotexe:
Weird, I can't repro that, it seems to work with `/` just fine (Discord does complain the first time you try and send it on the desktop client, but mobile clients have no such issue)

storytellermahkasad:
Same now, will adjust to warning and confirm. Not sure what was going on at the time :\ I blame Discord.

zhudotexe (approved):

PR #1532, opened 2021-06-08
Adds unified sheet import

storytellermahkasad:
### Summary
Adds a unified sheet importing that discovers the correct sheet type to parse with.

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/sheetManager.py
+++ b/cogs5e/sheetManager.py
@@ -538,14 +538,14 @@ class SheetManager(commands.Cog):
             url = beyond_url.group(1)
             parser = BeyondSheetParser(url)
         elif 'dicecloud.com' in url:
-            url = url.split('/character/')[-1].split('/')[0]
-            loading = await ctx.send('Loading character data from Dicecloud...')            
+            loading = await ctx.send('Loading character data from Dicecloud...')
+            url = url.split('/character/')[-1].split('/')[0]        
             prefix = 'dicecloud'
             parser = DicecloudParser(url)
         else:
             try:
-                url = extract_gsheet_id_from_url(url)
                 loading = await ctx.send('Loading character data from Google...')
+                url = extract_gsheet_id_from_url(url)
                 prefix = 'google'
                 parser = GoogleSheet(url)
             except ExternalImportError:
--- a/cogs5e/sheetManager.py
+++ b/cogs5e/sheetManager.py
@@ -502,6 +502,59 @@ class SheetManager(commands.Cog):
                 f"If you only wanted to update your character, run `{ctx.prefix}update` instead.")
         return True
 
+    @commands.command(name='import')
+    @commands.max_concurrency(1, BucketType.user)
+    async def import_sheet(self, ctx, url: str, *args):
+        """
+        Loads a character sheet in one of the accepted formats:
+            [Dicecloud](https://dicecloud.com/)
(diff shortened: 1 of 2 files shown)


cogs5e/sheetManager.py
```diff
@@ -502,6 +502,61 @@ async def _confirm_overwrite(self, ctx, _id):
...
+
+        __Sheet-specific Notes__
+        D&D Beyond:
+            Private sheets can be imported if you have linked your DDB and Discord accounts.  Otherwise, the sheet needs to be publicly shared.
+        Gsheet:
+            The sheet must be shared with Avrae for this to work.
+            Avrae's google account is `avrae-320@avrae-bot.iam.gserviceaccount.com`.
+
+        Dicecloud:
+            Share your character with `avrae` on Dicecloud (edit perms) for live updates.
```
zhudotexe, line 526:
```suggestion
            Share your character with `avrae` on Dicecloud (edit permissions) for live updates.
```

cogs5e/sheetManager.py
```diff
@@ -502,6 +502,61 @@ async def _confirm_overwrite(self, ctx, _id):
...
+        """
+        loading = await ctx.send('Loading character data to Avrae...')
+        prefix = ''
```
zhudotexe, line 529: What is this variable doing?

cogs5e/sheetManager.py
```diff
@@ -502,6 +502,61 @@ async def _confirm_overwrite(self, ctx, _id):
...
+        parser = None
+
+        url = await self._check_url(ctx, url) # check for < >
+        # Sheets in order: DDB, Dicecloud, Gsheet
+        beyond_url = DDB_URL_RE.match(url)
```
zhudotexe, line 534: so I know this is just copied from the respective existing import methods, but I think we can improve this, lol. I think the best way to do this is by parsing the url using [urllib](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) and checking the `netloc`

cogs5e/sheetManager.py
storytellermahkasad: It gets assigned per sheet type and is used in the confirm overwrites L554 but could refactor the block to not use it.

zhudotexe: Oh, there it is, just didn't see it

cogs5e/sheetManager.py
```diff
@@ -502,6 +502,61 @@ async def _confirm_overwrite(self, ctx, _id):
...
+        Gsheet:
+            The sheet must be shared with Avrae for this to work.
+            Avrae's google account is `avrae-320@avrae-bot.iam.gserviceaccount.com`.
+
+        Dicecloud:
+            Share your character with `avrae` on Dicecloud (edit permissions) for live updates.
+        """
+        loading = await ctx.send('Loading character data to Avrae...')
+        prefix = ''
+        parser = None
```
zhudotexe, line 530: These 3 lines can be removed (the first sends a duplicate message, and the latter two are guaranteed to be set in one of the if bodies)

zhudotexe: Although the message needs to be moved before `extract_gsheet_id_from_url` in the google case in this case

PR #1552, opened 2021-06-26
Adds (s)adv/(s)dis to effect

### LINKED ISSUE #432: AFR-253 make `adv/dis` arguments for `!i effect`
**Feature Request:** make `adv/dis` arguments for `!i effect`
**Who would use it?:** DMs, brewers, and combatants
**How would it work?:** `!init effect Something dis`
**Why should this be added?:** Makes it so penalties don't rely on the player to remember to do it

- R to the Ichie#6193

> Note - silverbass#2407
> 
> technically covered under afr-128 (#67)

> Upvote - Derixyleth#0636
> 
> Technically asked for in AFR-128 (#67), but still wanted, along with -criton and -d1 and -b1 and adv1 and dis1

> Note - GitHub - mommothazaz123
> 
> should adv/dis apply only to attacks?

> Note - silverbass#2407
> 
> adv/dis for attacks only, yes

> Note - Derixyleth#0636
> 
> adv/dis for attacks only, and possibly sadv/sdis for saves, but then that should probably be `-sadv <ability>` or `sdis <ability>` though.

> Upvote - Shrooblord#2735

> Upvote - Zekkle#9787

> Upvote - t̤̻͔͉͢h͕̩̪͚͙i̬̟̩̼͢g͓̼̠͜͟h̺̼̪̪̝#0001

Votes: +4 / -0

storytellermahkasad:
Resolves #432
### Summary
Adds `adv/dis` and `sadv/sdis <ability>` to `!i effect`. Both only activate if no other advantage args were passed in before the effects are resolved.

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/initTracker.py
+++ b/cogs5e/initTracker.py
@@ -921,6 +921,7 @@ class InitTracker(commands.Cog):
         -t <target> - Specifies more combatant's to target, chainable (e.g., "-t or1 -t or2").
         -parent <"[combatant]|[effect]"> - Sets a parent effect from a specified combatant.
         __Attacks__
+        adv/dis - Give advantage or disadvantage to the attack roll(s).
         -b <bonus> - Adds a bonus to hit.
         -d <damage> - Adds additional damage.
         -attack <"[hit]|[damage]|[description]"> - Adds an attack to the combatant. The effect name will be the name of the attack. No [hit] will autohit (e.g., -attack "|1d6[fire]|")
@@ -935,6 +936,7 @@ class InitTracker(commands.Cog):
         -ac <ac> - modifies ac temporarily; adds if starts with +/- or sets otherwise.
         -sb <save bonus> - Adds a bonus to all saving throws.
         -cb <check bonus> - Adds a bonus to all ability checks.
+        -sadv/sdis <ability> - Gives advantage/disadvantage on saving throws for the provided ability.
         -desc <description> - Adds a description of the effect."""
         combat = await Combat.from_ctx(ctx)
         args = argparse(args)
--- a/cogs5e/initTracker.py
+++ b/cogs5e/initTracker.py
@@ -936,8 +936,7 @@ class InitTracker(commands.Cog):
         -ac <ac> - modifies ac temporarily; adds if starts with +/- or sets otherwise.
         -sb <save bonus> - Adds a bonus to all saving throws.
         -cb <check bonus> - Adds a bonus to all ability checks.
-        sadv/sdis - Gives advantage/disadvantage on all saving throws.
-        -sadv/sdis <ability> - Gives advantage/disadvantage on saving throws for the provided ability.
+        -sadv/sdis <ability> - Gives advantage/disadvantage on saving throws for the provided ability, or "all" for all saves.
         -desc <description> - Adds a description of the effect."""
         combat = await Combat.from_ctx(ctx)
         args = argparse(args)
--- a/cogs5e/initTracker.py
+++ b/cogs5e/initTracker.py
@@ -921,7 +921,7 @@ class InitTracker(commands.Cog):
         -t <target> - Specifies more combatant's to target, chainable (e.g., "-t or1 -t or2").
         -parent <"[combatant]|[effect]"> - Sets a parent effect from a specified combatant.
         __Attacks__
-        adv/dis - Give advantage or disadvantage to the attack roll(s).
+        adv/dis - Give advantage or disadvantage to all attack rolls.
         -b <bonus> - Adds a bonus to hit.
         -d <damage> - Adds additional damage.
         -attack <"[hit]|[damage]|[description]"> - Adds an attack to the combatant. The effect name will be the name of the attack. No [hit] will autohit (e.g., -attack "|1d6[fire]|")
@@ -936,6 +936,7 @@ class InitTracker(commands.Cog):
         -ac <ac> - modifies ac temporarily; adds if starts with +/- or sets otherwise.
         -sb <save bonus> - Adds a bonus to all saving throws.
         -cb <check bonus> - Adds a bonus to all ability checks.
+        sadv/sdis - Gives advantage/disadvantage on all saving throws.
         -sadv/sdis <ability> - Gives advantage/disadvantage on saving throws for the provided ability.
         -desc <description> - Adds a description of the effect."""
         combat = await Combat.from_ctx(ctx)
--- a/cogs5e/models/automation/effects/attack.py
+++ b/cogs5e/models/automation/effects/attack.py
@@ -34,13 +34,7 @@ class Attack(Effect):
                                   "of a Target effect.")
 
         # arguments
-        args = autoctx.args
-        # adv/dis (#1552)
-        for check_arg in ['adv','dis']:
-            if autoctx.combatant.active_effects(check_arg):
-                args.update({check_arg: True})  # Because adv() only checks last() just forcibly add them
-
-        adv = args.adv(ea=True, ephem=True)
+        args = autoctx.args       
         crit = args.last('crit', None, bool, ephem=True) and 1
         nocrit = args.last('nocrit', default=False, type_=bool, ephem=True)
         hit = args.last('hit', None, bool, ephem=True) and 1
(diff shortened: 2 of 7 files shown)


storytellermahkasad:
Will need tests, wanted to see what others thought of implementation before adding them though.

cogs5e/models/initiative/effect.py
```diff
@@ -281,11 +283,24 @@ def parse_resist_str(resist_list):
...
 
 
+def parse_stat_choice(arg, _):
+    if not arg in STAT_ABBREVIATIONS:
+        raise InvalidArgument(f"{arg} is not a valid stat")
+    return arg
+
+
+def parse_stat_choice_str(stat):
+    return verbose_stat(stat)
```
zhudotexe, line 293: I think it's safe to just pass `verbose_stat` directly instead of making this wrapper

cogs5e/models/initiative/effect.py
```diff
@@ -281,11 +283,24 @@ def parse_resist_str(resist_list):
     return ', '.join([str(Resistance.from_dict(r)) for r in resist_list])
 
 
+def parse_stat_choice(arg, _):
+    if not arg in STAT_ABBREVIATIONS:
```
zhudotexe, line 287: nit: `not X in Y` should be `X not in Y` (https://www.flake8rules.com/rules/E713.html)

cogs5e/utils/checkutils.py
```diff
@@ -100,10 +100,18 @@ def run_save(save_key, caster, args, embed):
...
         embed.title = f'{caster.get_title_name()} makes {a_or_an(save_name)}!'
-
-    # ieffect -sb
+    
+    # ieffect handling
     if isinstance(caster, init.Combatant):
+        # -sb
         args['b'] = args.get('b') + caster.active_effects('sb')
+        # -sadv/sdis
+        if not args.adv(boolwise=False, ephem=True):
```
zhudotexe, line 109: Checking for the presence of advantage with `ephem=True` here would consume a use of any ephemeral argument. Is the purpose of this line to have any given adv/dis argument completely override effects rather than being mixed in?

zhudotexe: (If so, I believe `'adv' in args or 'dis' in args` would do the trick without consuming ephemeral uses)

cogs5e/utils/checkutils.py
```diff
@@ -100,10 +100,18 @@ def run_save(save_key, caster, args, embed):
...
+            adv = 0
```
zhudotexe, line 110: nit: unused variable

cogs5e/models/automation/effects/attack.py
```diff
@@ -54,14 +54,23 @@ def run(self, autoctx):
...
         if autoctx.combatant:
+            # bonus (#224)
             effect_b = '+'.join(autoctx.combatant.active_effects('b'))
             if effect_b and b:
                 b = f"{b}+{effect_b}"
             elif effect_b:
                 b = effect_b
 
+         # adv/dis (#1552)
+         # Only apply if we don't already have any advantage changes
```
zhudotexe, line 67: nit: check your comment indentation level!

zhudotexe (commented):
Overall, this looks great! Aside from the various nits, the main design question I have is why let a passed argument override the effect entirely?

Also, would you consider writing some tests? 😉

zhudotexe:
Oh, you'll probably need to add `sadv/sdis` handling code to the Save automation effect as well, since that doesn't go through checkutils

cogs5e/utils/checkutils.py
storytellermahkasad: Covered by force adding, given how argparse.adv() function only reads last.

storytellermahkasad:
> Oh, you'll probably need to add `sadv/sdis` handling code to the Save automation effect as well, since that doesn't go through checkutils

I can't for the life of me figure how to do this part, totally open to anyone's suggestions!
Narrowed it to `cogs5e\models\automation\effects\save.py` but can't find how I'd check effects for making sure it's the right save.

cogs5e/initTracker.py
```diff
@@ -921,6 +921,7 @@ async def effect(self, ctx, target_name: str, effect_name: str, *args):
         -t <target> - Specifies more combatant's to target, chainable (e.g., "-t or1 -t or2").
         -parent <"[combatant]|[effect]"> - Sets a parent effect from a specified combatant.
         __Attacks__
+        adv/dis - Give advantage or disadvantage to the attack roll(s).
```
zhudotexe, line 924:
```suggestion
        adv/dis - Give advantage or disadvantage to all attack rolls.
```

cogs5e/models/automation/effects/save.py
```diff
@@ -40,7 +40,15 @@ def run(self, autoctx):
...
         auto_fail = autoctx.args.last('fail', type_=bool, ephem=True)
         hide = autoctx.args.last('h', type_=bool)
-        adv = autoctx.args.adv(custom={'adv': 'sadv', 'dis': 'sdis'})
+
+        # check for combatant IEffects
+        if autoctx.combatant:
+            # Combine args/ieffect advantages - adv/dis (#1552)
+            adv = reconcile_adv(
+                adv= autoctx.args.last('sadv', bool, ephem=True) or autoctx.combatant.active_effects('adv'),
+                dis= autoctx.args.last('sdis', bool, ephem=True) or autoctx.combatant.active_effects('dis'))
```
zhudotexe, line 49: This doesn't seem to be checking that the save is for the correct ability, just that the argument in the ieffect exists (also, it's checking adv/dis instead of sadv/sdis)

cogs5e/models/initiative/effect.py
```diff
@@ -281,11 +283,24 @@ def parse_resist_str(resist_list):
...
 
 
+def parse_stat_choice(arg, _):
+    if arg not in STAT_ABBREVIATIONS:
+        raise InvalidArgument(f"{arg} is not a valid stat")
+    return arg
+
+
+def parse_stat_choice_str(stat):
+    return verbose_stat(stat)
```
zhudotexe, line 293:
```suggestion

```
we aren't using this anymore!

utils/functions.py
```diff
@@ -450,3 +450,23 @@ async def get_guild_member(guild, member_id):
...
+    :return: The combined advantage result
+    """
+    result = 0
+    if adv:
+        result += 1
+    if dis:
+        result += -1
+    if ea and not dis:
+        return 2
+    return {-1: False, 0: None, 1: True}.get(result)
```
zhudotexe, line 472: is returning boolwise correct here?

cogs5e/utils/checkutils.py
```diff
@@ -136,8 +138,15 @@ def _run_common(skill, args, embed, mod_override=None, rr_format="Check {}"):
...
-        # advantage
-        adv = args.adv(boolwise=True, ephem=True)
+        # ieffect handling
+        # Combine args/ieffect advantages - adv/dis (#1552)
+        if isinstance(caster, init.Combatant):
+            adv = reconcile_adv(
+                adv = args.last('adv', type_=bool, ephem=True) or stat in caster.active_effects('sadv'),
+                dis = args.last('dis', type_=bool, ephem=True) or stat in caster.active_effects('sdis'))
+        else:
+            adv = args.adv(boolwise=True, ephem=True)
```
zhudotexe, line 148: From Discord:
> :\ So _run_common takes in the full args but not the context. Can't think of a way to do the reconcile there, which means it needs to be in run_save instead.
> Since the args get dropped on the floor after _run_common, is it okay to just insert them like they are now?
> Wait- no it doesn't. Dangit.
>The easiest answer I can think of is passing the caster down one more level.

Well, we're already modifying args in L108... you're right that the args don't get reused in outer scopes on checks, so I think modifying args is fine. It seems a little more elegant than having the `stat` argument be kind of ambiguous and having to pass the caster down a level, anyway

utils/functions.py
storytellermahkasad: It's not, I think I put it back while we were figuring out the better version where you pass in all the args. Will remove.

cogs5e/utils/checkutils.py
storytellermahkasad: Got it will return to the version that just appends as it's much simpler.

zhudotexe (approved):

PR #1554, opened 2021-06-27
Allows newlines in multiline

storytellermahkasad:
### Summary
A possibly poorly thought through implementation of adding newlines to multiline.

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogsmisc/customization.py
+++ b/cogsmisc/customization.py
@@ -472,8 +472,9 @@ class Customization(commands.Cog):
     @commands.command()
     @commands.max_concurrency(1, BucketType.user)
     async def multiline(self, ctx, *, cmds: str):
-        """Runs each line as a separate command, with a 1 second delay between commands.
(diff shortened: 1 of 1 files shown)


cogsmisc/customization.py
```diff
@@ -444,9 +444,10 @@ async def multiline(self, ctx, *, cmds: str):
         !spell Fly
         !monster Rat"
         """
-        cmds = cmds.splitlines()
+        # Remove the first prefix to simplify loop. Split only on actual new commands
+        cmds = cmds.lstrip(ctx.prefix).split(f"\n{ctx.prefix}")
```
zhudotexe, line 448: If only we were in Python 3.9! `.removeprefix` is more appropriate here than `.lstrip` (for example, if my prefix were `a!`, and the command were `a!attack`, this would strip the `a` from `attack`).

PR #1556, opened 2021-06-30
Ignore crit state on children of Save Automation

storytellermahkasad:
### Summary
~~Adds handling for nocritical that pairs with underlying nocrit.~~
Ignores crit state for children of Save automation, restoring it afterwards. Allowing a damage's dice to not be doubled on critical. Mostly intended for saves as part of attacks. (Snake poison, etc)

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/models/automation/effects/save.py
+++ b/cogs5e/models/automation/effects/save.py
@@ -101,15 +101,14 @@ class Save(Effect):
             is_success = False
 
         # Disable critical damage state for children #1556
-        old_no_crit = autoctx.args.last('nocrit', default=False, type_=bool)
-        autoctx.args.update({'nocrit': True})
+        autoctx.in_save = True
 
         if is_success:
             children = self.on_success(autoctx)
         else:
             children = self.on_fail(autoctx)
-        
-        autoctx.args.update({'nocrit': old_no_crit})  # Restore proper crit state #1556
+
+        autoctx.in_save = False  # Restore proper crit state #1556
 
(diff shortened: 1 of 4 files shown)


storytellermahkasad:
Likely to be modified to a different method. Will close if needed, otherwise changes implemented here.

cogs5e/models/automation/effects/save.py
```diff
@@ -96,13 +97,20 @@ def run(self, autoctx):
                     autoctx.add_pm(str(autoctx.ctx.author.id), out)
                     autoctx.queue(f"**{save_blurb}**: 1d20...{success_str}")
         else:
-            autoctx.meta_queue('{} Save'.format(save_skill[:3].upper()))
+            autoctx.meta_queue('{} Save'.format(stat.upper()))
             is_success = False
 
+        # Disable critical damage state for children #1556
+        old_no_crit = autoctx.args.last('nocrit', default=False, type_=bool)
+        autoctx.args.update({'nocrit': True})
```
zhudotexe, line 105: Why update the arguments rather than the internal crit state (ala `Attack.on_crit()`)? If, for example, some automation existed as such:
```yaml
- type: save
  save: ...
  fail:
    - type: attack
      hit:
        - type: damage
          damage: ...
```
the inner damage should still crit if the inner attack crits, but this approach would prevent that from happening.

zhudotexe (approved):

PR #1568, opened 2021-07-08
Add SimpleCombat Metadata

storytellermahkasad:
### Summary
Adds a metadata dict to SimpleCombat, useful for storing info about the combat without a placeholder.

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/aliasing/api/combat.py
+++ b/aliasing/api/combat.py
@@ -54,23 +54,23 @@ class SimpleCombat:
         return None
 
     @property
-    def note(self):
+    def info(self):
         """
-        The description on the combatant. ``None`` if not set.
+        Information about the combat. ``None`` if not set.
 
         :rtype: str or None
         """
-        return self._combat.notes
+        return self._combat.info
 
-    def set_note(self, note: str):
+    def set_info(self, info: str):
         """
-        Sets the combatant's note.
+        Sets the combat's information.
 
-        :param str note: The new note.
+        :param str note: The new info.
         """
-        if note is not None:
-            note = str(note)
-        self._combat.notes = note
+        if info is not None:
+            info = str(info)
+        self._combat.info = info
 
     def get_group(self, name):
         """
--- a/aliasing/api/combat.py
+++ b/aliasing/api/combat.py
@@ -11,6 +11,7 @@ from utils.argparser import ParsedArguments
 class SimpleCombat:
     def __init__(self, combat, me):
         self._combat: Combat = combat
+        self.description = None
 
         self.combatants = [SimpleCombatant(c) for c in combat.get_combatants()]
         self.groups = [SimpleGroup(c) for c in combat.get_groups()]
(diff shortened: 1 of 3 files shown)


aliasing/api/combat.py
```diff
@@ -67,6 +70,51 @@ def get_group(self, name):
...
+        :param str k: The metadata key to set
+        :param str v: The metadata value to set
+
+        Example:
+
+        >>> set_metadata("Test", dump_json({"Status": ["Mario", 1, 2]}))
+        """
+        if sum(len(k + v) for k, v in self._combat._metadata.items()) > MAX_METADATA_SIZE:
+            raise ValueError
+        self._combat._metadata[k] = v
```
zhudotexe, line 87:
```suggestion
        self._combat._metadata[str(k)] = str(v)
        if sum(len(k + v) for k, v in self._combat._metadata.items()) > MAX_METADATA_SIZE:
            raise ValueError("Combat metadata is too large")
```
- Make sure to do the assignment before size checking so it's actually checking the size of the resulting dict
- Cast key, value to str just in case
- ValueError needs a message

aliasing/api/combat.py
```diff
@@ -67,6 +70,51 @@ def get_group(self, name):
...
+
+        :param str k: The metadata key to get
+        :param default: What to return if the name is not set.
+
+        Example:
+
+        >>> get_metadata("Test")
+        "{"Status": ["Mario", 1, 2]}"
+        """
+        return self._combat._metadata.get(k, default)
```
zhudotexe, line 101:
```suggestion
        '{"Status": ["Mario", 1, 2]}'
        """
        return self._combat._metadata.get(str(k), default)
```

aliasing/api/combat.py
```diff
@@ -67,6 +70,51 @@ def get_group(self, name):
...
+        :param str k: The metadata key to remove
+        :return: The removed value or ``None`` if the key is not found.
+        :rtype: str or None
+
+        Example:
+
+        >>> delete_metadata("Test")
+        "{"Status": ["Mario", 1, 2]}"
+        """
+        return self._combat._metadata.pop(k, None)
```
zhudotexe, line 116:
```suggestion
        '{"Status": ["Mario", 1, 2]}'
        """
        return self._combat._metadata.pop(str(k), None)
```

cogs5e/initTracker.py
```diff
@@ -592,6 +592,18 @@ async def note(self, ctx, name: str, *, note: str = ''):
...
+        """Attaches information to the combat."""
+        combat = await Combat.from_ctx(ctx)
+
+        combat.info = info
+        if info == '':
+            await ctx.send("Removed info.")
+        else:
+            await ctx.send("Added info.")
+        await combat.final()
+
```
zhudotexe, line 606: This should be removed

cogs5e/models/initiative/combat.py
```diff
@@ -32,6 +33,9 @@ def __init__(self, channel_id, message_id, dm_id, options, ctx,
         self._turn = turn_num
         self._current_index = current_index
         self.ctx = ctx
+        if metadata is None:
+            metadata = {}
```
zhudotexe, line 37: Convention is to put all mutable default arguments at the top (in this case, right under `if combatants is None`)

PR #1572, opened 2021-07-10
Return group.name instead of group_id

storytellermahkasad:
### Summary
SimpleCombatant.group is returning Id instead of Name. Now returns the name or none as per spec.

#### PR Type

- [X] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/models/initiative/combat.py
+++ b/cogs5e/models/initiative/combat.py
@@ -279,7 +279,7 @@ class Combat:
         :param strict: Whether group name must be a full case insensitive match.
         :return: The combatant group.
         """
-        if name: # catch no group without having to loop
+        if name:  # catch no group without having to loop
             if name in self._combatant_id_map and isinstance(self._combatant_id_map[name], CombatantGroup):
                 return self._combatant_id_map[name]
             if strict:
--- a/cogs5e/models/initiative/combat.py
+++ b/cogs5e/models/initiative/combat.py
@@ -279,17 +279,16 @@ class Combat:
         :param strict: Whether group name must be a full case insensitive match.
         :return: The combatant group.
         """
-        if name:  # catch no group without having to loop
-            if name in self._combatant_id_map and isinstance(self._combatant_id_map[name], CombatantGroup):
-                return self._combatant_id_map[name]
-            if strict:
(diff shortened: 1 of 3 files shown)


cogs5e/models/initiative/combat.py
```diff
@@ -279,16 +279,17 @@ def get_group(self, name, create=None, strict=True):
         :param strict: Whether group name must be a full case insensitive match.
         :return: The combatant group.
         """
-        if name in self._combatant_id_map and isinstance(self._combatant_id_map[name], CombatantGroup):
-            return self._combatant_id_map[name]
-        if strict:
-            grp = next((g for g in self.get_groups() if g.name.lower() == name.lower()), None)
-        else:
-            grp = next((g for g in self.get_groups() if name.lower() in g.name.lower()), None)
+        if name:  # catch no group without having to loop
```
zhudotexe, line 282: What case spurred this change, out of curiosity?

zhudotexe: Also, to prevent having to indent the rest of the method, it's cleaner to do 
```py
if not name:
    return None
```
(in fact, I think this raises an UnboundLocalError, since `grp` is undefined when it reaches the return)

zhudotexe: Although, an empty name might be better off raising something, since it's not really a valid group name anyway, and just returning None might have unexpected effects when `create` is set

zhudotexe: Although the init system does let group names be empty so it might be better to just eat the loops here

zhudotexe: Oh, I see, `name` might be `None` from `Combatant.get_group_name` - I think that check should be done in the combatant's method, then

cogs5e/models/initiative/combatant.py
```diff
@@ -251,6 +251,10 @@ def set_group(self, group_name):
                 self.combat.goto_turn(self, True)
             return c_group
 
+    def get_group_name(self):
```
zhudotexe, line 254: Consider making this `get_group` to be more general and using `get_group().name` (with relevant None checks) in the AliasCombatant method

zhudotexe (approved):

PR #1581, opened 2021-07-16
Counters Set lastCounterName on i

storytellermahkasad:
### Summary
Resolves #AVR-810 (Maybe)
If resources ignored, still set counter name from passed counter.

#### PR Type

- [X] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/models/automation/effects/usecounter.py
+++ b/cogs5e/models/automation/effects/usecounter.py
@@ -73,21 +73,17 @@ class UseCounter(Effect):
         autoctx.metavars['lastCounterUsedAmount'] = result.used_amount
         return result
 
-    def get_and_use_counter(self, autoctx, amount, ignore_resources: bool = False):  # this is not in run() because indentation
-        if ignore_resources:
+    def get_and_use_counter(self, autoctx, amount, ignore_resources: bool = False):
+        if ignore_resources:  # handled here to return counter name (#1582)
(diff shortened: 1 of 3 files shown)


storytellermahkasad:
Should this be applied only on the spell version, or should the documentation be updated:
`lastCounterName (str)` The name of the last used counter. If it was a spell slot, the level of the slot (safe to cast to int, i.e. `int(lastCounterName)`). (`None` on error).

Perhaps as:
`lastCounterName (str)` The name of the last used counter. If it was a spell slot, the level of the slot (safe to cast to int, i.e. `int(lastCounterName)`). (Will always be set if `i` (ignore resources) is passed. `None` on error).

cogs5e/models/automation/effects/usecounter.py
```diff
@@ -77,7 +73,12 @@ def run(self, autoctx):
         autoctx.metavars['lastCounterUsedAmount'] = result.used_amount
         return result
 
-    def get_and_use_counter(self, autoctx, amount):  # this is not in run() because indentation
+    def get_and_use_counter(self, autoctx, amount, ignore_resources: bool = False):  # this is not in run() because indentation
+        if ignore_resources:
+            return UseCounterResult(counter_name=self.counter.name,
```
zhudotexe, line 78: This is an AttributeError since the type of `self.counter` here is `str | AbilityReference`

PR #1585, opened 2021-07-17
Let effects override innate resists

storytellermahkasad:
### Summary
Right now effects on a combatant can't override innate resistances if they are of the same type as an existing resist
Ex: `-neutral necrotic` on a creature with Resistance: necrotic

#### PR Type

- [X] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/models/initiative/combatant.py
+++ b/cogs5e/models/initiative/combatant.py
@@ -175,7 +175,7 @@ class Combatant(BaseCombatant, StatBlock):
     @property
     def resistances(self):
         out = self._resistances.copy()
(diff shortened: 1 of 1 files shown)


Croebh:
To be clear, it works fine on `!attack` and `!cast`, but it breaks on `SimpleCombatant.damage()`

zhudotexe:
The `overwrite` kwarg removes any base resistance, which might not be desirable (e.g. a monster with a base resistance to fire, but some effect gives them fire vuln - without `overwrite` it will have both resistance and vuln, but with `overwrite` just vuln)

PR #1589, opened 2021-07-21
Prevent error on empty -thumb image

storytellermahkasad:
### Summary
Resolves #AVR-811
Prevents error on empty `-thumb <image>` by returning blank if the var doesn't exist

#### PR Type

- [X] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/models/character.py
+++ b/cogs5e/models/character.py
@@ -56,7 +56,7 @@ class Character(StatBlock):
 
         # main character info
         self._description = description
-        self._image = image or ''
+        self._image = image
 
         # customization
         self.cvars = cvars
@@ -194,7 +194,7 @@ class Character(StatBlock):
 
     @property
     def image(self):
-        return self.overrides.image or self._image
+        return self.overrides.image or self._image or ''
 
     # ---------- CSETTINGS ----------
     def get_setting(self, setting, default=None):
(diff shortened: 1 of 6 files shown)


driesceuppens:
I just checked and the same problem seems to happen with `-image <image>` so you could change that right away too

cogs5e/models/character.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, owner: str, upstream: str, active: bool, sheet_type: str, imp
 
         # main character info
         self._description = description
-        self._image = image
+        self._image = image or ''
```
zhudotexe, line 59: Instead of modifying the character data, a better pattern is to put this in the `image` property getter

cogs5e/pbpUtils.py
```diff
@@ -54,7 +54,7 @@ async def embed(self, ctx, *, args):
         args = argparse(args)
         embed.title = args.last('title')
         embed.description = args.last('desc')
-        embed.set_thumbnail(url=args.last('thumb', '') if 'http' in str(args.last('thumb')) else '')
+        embed.set_thumbnail(url=maybe_http_url(args.last('thumb', '')))
         embed.set_image(url=args.last('image', '') if 'http' in str(args.last('image')) else '')
```
zhudotexe, line 58: Seems like this would be a good place to use `maybe_http_url` too

zhudotexe (approved):

PR #1635, opened 2021-10-06
Update footer for `!alias list` and variants

Croebh:
### Summary
Currently the footer in `!alias` says `!alias <page>` when ran as just `!alias`, but properly says `!alias list <page>` when ran with `!alias list`, and the same for snippets and the server variants of each. This PR changes it to save the base command and display that instead.

#### PR Type

- [x] This PR fixes an issue.
#### Other


Croebh:
Edited to reflect the fact this PR also resolves #1638

PR #1639, opened 2021-10-10
Add a message telling users to look closely at their linked discord account when running `!ddb`

Croebh:
### Summary
One of most common issues when linking your D&D Beyond accounts with Avrae is that Discord will actually link a 'proxy' account. Unfortunately, this is not at all obvious to the end user, and to them it feels as though the accounts should be linked, but aren't. There is currently no wording in the `!ddb` command or on the D&D Beyond website (where you link) that this is even a possibility, which has lead to some very confused users asked what was going on. This PR adds a message to the `!ddb` command when your accounts are not detect as linked.

Resolves AFR-881

#### PR Type

- [ ] This PR fixes an issue.
#### Other


Croebh:
Gdi I forgot to make the branch based on the master >_>

PR #1661, opened 2021-11-08
Remove vague error if user forgets quotes on spells containing 'with'

Croebh:
### Summary
Adds an additional check to make sure `with_arg` isnt `"True"`, which is the case if the caster forgot quotes when casting a spell containing `with` in the name, such as `!cast speak with dead`. Without this change, it gives the user an error: `True is not a valid stat to cast with.`, which is incredibly unhelpful when you aren't actually trying to use the `-with` argument,

#### PR Type

- [x] This PR fixes an issue.
#### Other


zhudotexe (approved):
copying some thoughts from Discord - while this PR LGTM, it's very specific

> what if we added an arg that was ParsedArguments.get(..., ignore_true=True) or something
that just straight up ignored any value that was the literal True
or even set that to the default and added a .contains or just double checking that arg in ParsedArguments is used for when we actually use True
that would fix things like -t d2 (but not -t ea, alas)

PR #1679, opened 2021-12-04
Parse automation YAMLs in CritterDB descriptions

Lathaon:
### Summary
Allows Avrae to parse YAMLs within `<avrae hidden></avrae>` tags in CritterDB action descriptions. Old override patterns still work too.

It appears to work but I am new to this so would appreciate more experienced people checking/improving the code.

I prepared a test monster with one attack that isn't overridden, one that's overridden using the old format, one that's overridden with a JSON and one that's overridden with a YAML.
https://critterdb.com/#/publishedbestiary/view/61ab7827e6e6ad7f5718c354

FR got 16 upvotes but I don't think it's still open. https://discord.com/channels/269275778867396608/834514549356625960/834970483148587048

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/models/homebrew/bestiary.py
+++ b/cogs5e/models/homebrew/bestiary.py
@@ -330,6 +330,14 @@ def _monster_factory(data, bestiary_name):
     legactions, atks = parse_critterdb_traits(data, 'legendaryActions')
     attacks.extend(atks)
 
+    name_duplications = {}
+    for atk in attacks:
+        if atk['name'] in name_duplications:
+            name_duplications[atk['name']] += 1
+            atk['name'] = atk['name'] + str(name_duplications[atk['name']])
+        else:
+            name_duplications[atk['name']] = 1
+
     attacks = AttackList.from_dict(attacks)
     spellcasting = parse_critterdb_spellcasting(traits, ability_scores)
 
@@ -369,13 +377,13 @@ def parse_critterdb_traits(data, key):
                     attack_yaml = yaml.safe_load(override.group(1))
                     if not isinstance(attack_yaml, list):
                         attack_yaml = [attack_yaml]
-                    if not all(isinstance(x, dict) for x in attack_yaml):
-                        raise ValueError('Invalid YAML')
-                    conflicts = [a for a in attacks if a['name'].lower() in [new['name'].lower() for new in attack_yaml]]
-                    for a in conflicts:
-                        attacks.remove(a)
+                    for atk in attack_yaml:
+                        if not isinstance(atk, dict):
+                            raise ValueError('Invalid YAML')
+                        elif 'name' not in atk:
+                            atk['name'] = name
                     attacks.extend(attack_yaml)
-                except (yaml.YAMLError, ValueError, KeyError):
+                except (yaml.YAMLError, ValueError):
                     simple_override = AVRAE_ATTACK_OVERRIDES_SIMPLE_RE.fullmatch(override.group(1))
                     if simple_override:
                         attacks.append({'name': simple_override.group(1) or name,
--- a/cogs5e/models/homebrew/bestiary.py
+++ b/cogs5e/models/homebrew/bestiary.py
@@ -373,7 +373,7 @@ def parse_critterdb_traits(data, key):
                     for a in conflicts:
                         attacks.remove(a)
                     attacks.extend(attack_yaml)
-                except Exception:
+                except yaml.YAMLError:
                     simple_override = AVRAE_ATTACK_OVERRIDES_SIMPLE_RE.fullmatch(override.group(1))
                     if simple_override:
                         attacks.append({'name': simple_override.group(1) or name,
--- a/cogs5e/models/homebrew/bestiary.py
+++ b/cogs5e/models/homebrew/bestiary.py
@@ -392,7 +392,7 @@ def parse_critterdb_traits(data, key):
                             atk['name'] = atk_name = atk.get('name') or name
                             try:
                                 attacks.append(Attack.from_dict(atk))
-                            except KeyError:
+                            except Exception:
                                 raise ExternalImportError(
                                     f"An automation YAML contained an invalid attack ({data['name']}: {atk_name})")
                         else:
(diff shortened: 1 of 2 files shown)


cogs5e/models/homebrew/bestiary.py
```diff
@@ -362,9 +365,20 @@ def parse_critterdb_traits(data, key):
...
-                                'details': desc})
+                try:
+                    attack_yaml = yaml.safe_load(override.group(1))
+                    if not isinstance(attack_yaml, list):
+                        attack_yaml = [attack_yaml]
+                    conflicts = [a for a in attacks if a['name'].lower() in [new['name'].lower() for new in attack_yaml]]
+                    for a in conflicts:
+                        attacks.remove(a)
+                    attacks.extend(attack_yaml)
+                except Exception:
```
1drturtle, line 376: `except Exception` is a very broad exception, maybe make this more specific (ex: `yaml.YAMLError`)

Lathaon: Done.

I initially had it running `AttackList.from_dict(attack_yaml)` just to check early on if it was a valid attack, and stole Exception from sheetManager.py, but it seemed unnecessary...

cogs5e/models/homebrew/bestiary.py
```diff
@@ -362,9 +365,22 @@ def parse_critterdb_traits(data, key):
...
-                                'details': desc})
+                try:
+                    attack_yaml = yaml.safe_load(override.group(1))
+                    if not isinstance(attack_yaml, list):
+                        attack_yaml = [attack_yaml]
+                    if not all(isinstance(x, dict) for x in attack_yaml):
+                        raise ValueError('Invalid YAML')
+                    conflicts = [a for a in attacks if a['name'].lower() in [new['name'].lower() for new in attack_yaml]]
+                    for a in conflicts:
+                        attacks.remove(a)
```
Croebh, line 376: So I believe standard for normal attack lists for PC's would be to append a number to the 'duplicate' attacks, so like Longsword, Longsword2, Longsword3.

You could also account for them missing the name key by grabbing the name of the action, which is I believe the default for basic attack overrides atm

Lathaon: On the first point:

The current system actually has no conflict detection to stop you giving a CritterDB creature multiple attacks with the same name. I think it's a very different context from character sheet imports (a basic and widely used feature) but this does seem like an improvement on how it currently works so I've added it in.

On the second point:

Makes sense. I've added this in.

zhudotexe:
Seems solid to me. How does it handle attack validation?

Lathaon:
> 

You raise a good point.

If it fails to read it as YAML, or if it succeeds but the outermost structure of the YAML is not a dict, then it will attempt to read it as the old format and then skip that attack if that doesn't fit either,

However, if the YAML satisfies those requirements but is still not a valid attack, then the import can fail with a rather unhelpful "Uh oh, that wasn't supposed to happen!" when it tries to do `attacks = AttackList.from_dict(attacks)`. I got a KeyError from changing a `-type: save` to a `- type: dinosaur` for example.

I think it would be best to catch this early, right after successfully loading a YAML. But I have a couple of questions:

1. Is there a simple but cheap way to check if it's a valid attack, using existing functions?
2. Should the attack then be skipped, or should the import still fail but with a better error message that hopefully helps the user to identify the problem? I suppose this question applies to the existing checks as well.

zhudotexe:
sheetManagerL160 simply catches the exception from `AttackList.form_dict()` - which while probably good *enough*, has never really been a check I've been satisfied with. The right solution here would be to reimplement the Attack validation found in the API (or making an API call to perform this validation), although I'm not certain if that's within the scope of this PR.

Lathaon:
I agree it seems out of scope, so will leave that for a separate PR hopefully made by someone with a better understanding of what's required.

For now, I've added a less ideal, but seemingly working solution.

cogs5e/models/homebrew/bestiary.py
```diff
@@ -382,8 +383,9 @@ def parse_critterdb_traits(data, key):
                             raise ValueError('Invalid YAML')
                         elif 'name' not in atk:
                             atk['name'] = name
+                        Attack.from_dict(deepcopy(atk))
```
zhudotexe, line 386: Why is the deepcopy necessary here?

cogs5e/models/homebrew/bestiary.py
```diff
@@ -362,33 +372,73 @@ def parse_critterdb_traits(data, key):
...
         if overrides:
             for override in overrides:
-                attacks.append({'name': override.group(1) or name,
-                                'attackBonus': override.group(2) or None, 'damage': override.group(3) or None,
-                                'details': desc})
+                if override.group('simple'):
+                    attacks.append(Attack.from_dict({
+                        'name': override.group(2) or name,
+                        'attackBonus': override.group(3) or None,
+                        'damage': override.group(4) or None,
```
zhudotexe, line 379: With the new regex groups, check to make sure the group numbers haven't changed

cogs5e/models/homebrew/bestiary.py
```diff
@@ -362,33 +372,73 @@ def parse_critterdb_traits(data, key):
...
+                        raise ExternalImportError(
+                            f"Monster had an invalid automation YAML ({data['name']}: {name})")
+                    if not isinstance(attack_yaml, list):
+                        attack_yaml = [attack_yaml]
+                    for atk in attack_yaml:
+                        if isinstance(atk, dict):
+                            atk['name'] = atk_name = atk.get('name') or name
+                            try:
+                                attacks.append(Attack.from_dict(atk))
+                            except KeyError:
```
zhudotexe, line 395: This exception bound may not catch all possible exceptions

cogs5e/models/homebrew/bestiary.py
Lathaon: They did, but these numbers are the correct ones.

cogs5e/models/homebrew/bestiary.py
Lathaon: Noted, have changed this to `except Exception:` in the latest commit. Thanks

cogs5e/models/homebrew/bestiary.py
Lathaon: Removed in later version. Was necessary at the time because Attack.from_dict() was being run twice and has destructive popping.

zhudotexe (approved):

PR #1704, opened 2022-02-01
Adds the targets health to the search and select combatant embed

Croebh:
### Summary
This just adds the targets hp_str() to the search and select embed that comes up with users target ambiguously. This just gives a little bit more context to the selection.

#### PR Type

- [ ] This PR fixes an issue.
#### Other


cogs5e/initiative/combat.py
```diff
@@ -383,7 +383,7 @@ async def select_combatant(self, name, choice_message=None, select_group=False):
         :return: The selected Combatant, or None if the search failed.
         """
         return await search_and_select(
-            self.ctx, self.get_combatants(select_group), name, lambda c: c.name,
+            self.ctx, self.get_combatants(select_group), name, lambda c: f"{c.name} {c.hp_str()}",
             message=choice_message
```
zhudotexe, line 387:
```suggestion
            self.ctx, self.get_combatants(select_group), name, lambda c: c.name,
            message=choice_message, selectkey=lambda c: f"{c.name} {c.hp_str()}
```
Changing the `key` parameter actually changes what the searcher matches on - in the failed test, this causes an explicit match against the name `KO1` to instead open a prompt with partial matches on `KO1 - <5/5 HP>` and `KO10 - <5/5 HP>`. Instead, the `selectkey` parameter is used to display a different name in the select step without touching the key the searcher uses to match.

Croebh: Oh, that makes a lot more sense

PR #1707, opened 2022-02-06
Allow HP-sorted targeting in automation

### LINKED ISSUE #1710: AFR-912 A version of automation target node like all/each but iterating over the targets in order of lowest to highest HP
**Feature Request:** A version of automation target node like all/each but iterating over the targets in order of lowest to highest HP
**Who would use it?** Most likely Croebh writing the automation, but would benefit players using Sleep and Color Spray (both available to starting characters)
**What part of the bot?** Automation
**How would it work?** Instead of `target: each` you'd choose `target: by hp` or something
**Why should this be added?** It would allow automation of Sleep and Color Spray, which apply their effects in hp order, without the need for aliases. There might be other effects like this, but those were the ones that came to mind.

- Lathaon#6649

> Upvote - Lathaon#6649

> Upvote - Croebh#5603

> Upvote - TheReverendB#1377

> Upvote - Astral Observatory#3378

> Upvote - Asslo#8135

> Upvote - Dr Turtle#1771

> Upvote - mayaaaluna#4103

Lathaon:
### Summary
Resolves #1710 AFR-912

Allows target nodes to be sorted prior to execution, so we can have automation for spells like Sleep and Color Spray.
Currently the only recognised options are `sort_by: hp_asc` and `sort_by: hp_desc` but this could easily be extended.
Adds a `sort_by` key to the target node, so it can be used for positional target nodes as well as each/all.

#### PR Type

- [ ] This PR fixes an issue.
#### Other


cogs5e/models/automation/effects/target.py
```diff
@@ -27,8 +30,23 @@ def run(self, autoctx):
...
+                sort_last = float('-inf')
+            else:
+                descending = False
+                sort_last = float('inf')
+
+            if self.sort_by in ('hp_asc', 'hp_desc'):
+                targets = sorted(
+                    autoctx.targets, reverse=descending,
+                    key=lambda t: (t.hp or sort_last) if isinstance(t, StatBlock) else sort_last
+                )
```
zhudotexe, line 46:
```suggestion
        if self.sort_by == 'hp_asc':
            targets = sorted(
                autoctx.targets, 
                key=lambda t: (t.hp or float('inf')) if isinstance(t, StatBlock) else float('inf')
            )
        elif self.sort_by == 'hp_desc':
            targets = sorted(
                autoctx.targets,
                key=lambda t: (t.hp or float('-inf')) if isinstance(t, StatBlock) else float('-inf'),
                reverse=True
            )
```
Clever, but it's actually shorter and more readable to just do a simple if-switch 😛 (and maybe more so if you pull out the `key` into a function in utils, like `target_hp_or_default(target, default)` - this would also let you get rid of the nasty edge case where a `None` hp combatant could be sorted in front of a `0` hp combatant as both `None` and `0` are falsy)

Lathaon: Fair. I have made these changes, including the function, but can't test them until I get back home. Good catch with the `0`

zhudotexe (approved):

PR #1713, opened 2022-02-12
Add cc edit for modifying counters in place

storytellermahkasad:
### Summary
Adds `cc edit` for modifying a counter in place. Will attempt to keep the current counter value, clamping it to the new min/max as needed. Allows for using `none` to remove an argument from the counter, still uses all validation rules of `cc create`.

Resolves #AFR-863

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/aliasing/api/character.py
+++ b/aliasing/api/character.py
@@ -171,7 +171,7 @@ class AliasCharacter(AliasStatBlock):
         """
         Edits an existing custom counter.
 
-        Pass `none` to remove an argument entirely.
+        Pass ``None`` to remove an argument entirely.
         Will clamp counter value to new limits if needed.
 
         :param str name: The name of the counter to edit.
--- a/aliasing/api/character.py
+++ b/aliasing/api/character.py
@@ -2,7 +2,7 @@ from functools import cached_property
 
 import cogs5e.models.sheet.player as player_api
 from aliasing import helpers
-from aliasing.utils import optional_cast_arg_or_default
+from aliasing.utils import optional_cast_arg_or_default, UNSET
 from aliasing.api.statblock import AliasStatBlock
 from cogs5e.models.errors import ConsumableException
 
@@ -165,9 +165,9 @@ class AliasCharacter(AliasStatBlock):
             self.delete_cc(name)
         return self.create_cc_nx(name, *args, **kwargs)
 
-    def edit_cc(self, name: str, minVal: str = None, maxVal: str = None, reset: str = None,
-                dispType: str = None, reset_to: str = None, reset_by: str = None,
-                title: str = None, desc: str = None):
+    def edit_cc(self, name: str, minVal: str = UNSET, maxVal: str = UNSET, reset: str = UNSET,
+                dispType: str = UNSET, reset_to: str = UNSET, reset_by: str = UNSET,
+                title: str = UNSET, desc: str = UNSET):
         """
         Edits an existing custom counter.
 
@@ -198,15 +198,14 @@ class AliasCharacter(AliasStatBlock):
         title = optional_cast_arg_or_default(title, default=counter.title)
         desc = optional_cast_arg_or_default(desc, default=counter.desc)
 
-        if self.cc_exists(name):
-            edit_consumable = player_api.CustomCounter.new(
-                self._character, name, minVal, maxVal, reset, dispType,
-                title=title, desc=desc, reset_to=reset_to, reset_by=reset_by)
-            edit_consumable.set(counter.value)
-            self._character.consumables.insert(self._character.consumables.index(counter), edit_consumable)
-            self._character.consumables.remove(counter)
-            self._consumables = None  # reset cache
-            return AliasCustomCounter(edit_consumable)
+        edit_consumable = player_api.CustomCounter.new(
+            self._character, name, minVal, maxVal, reset, dispType,
+            title=title, desc=desc, reset_to=reset_to, reset_by=reset_by)
+        edit_consumable.set(counter.value)
+        self._character.consumables.insert(self._character.consumables.index(counter), edit_consumable)
+        self._character.consumables.remove(counter)
+        self._consumables = None  # reset cache
+        return AliasCustomCounter(edit_consumable)
 
     def cc_exists(self, name):
         """
--- a/aliasing/api/character.py
+++ b/aliasing/api/character.py
@@ -163,6 +163,79 @@ class AliasCharacter(AliasStatBlock):
             self.delete_cc(name)
         return self.create_cc_nx(name, *args, **kwargs)
 
+    def edit_cc(self, name: str, minVal: str = None, maxVal: str = None, reset: str = None,
+                dispType: str = None, reset_to: str = None, reset_by: str = None,
+                title: str = None, desc: str = None):
+        """
+        Edits an existing custom counter.
+
+        Pass `none` to remove an argument entirely.
+        Will clamp counter value to new limits if needed.
+
+        :param str name: The name of the counter to edit.
+        :param str minVal: The minimum value of the counter. Supports :ref:`cvar-table` parsing.
+        :param str maxVal: The maximum value of the counter. Supports :ref:`cvar-table` parsing.
+        :param str reset: One of ``'short'``, ``'long'``, ``'hp'``, ``'none'``, or ``None``.
+        :param str dispType: Either ``None`` or ``'bubble'``.
+        :param str reset_to: The value the counter should reset to. Supports :ref:`cvar-table` parsing.
+        :param str reset_by: How much the counter should change by on a reset. Supports dice but not cvars.
+        :param str title: The title of the counter.
+        :param str desc: The description of the counter.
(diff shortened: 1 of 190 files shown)


storytellermahkasad:
CC Edit Help

cogs5e/gametrack.py
```diff
@@ -584,6 +584,60 @@ async def customcounter_create(self, ctx, name, *args):
...
+        e_reset = args.last('reset', counter.reset_on)
+        e_max = args.last('max', counter.get_max())
+        e_min = args.last('min', counter.get_min())
+        e_type = args.last('type', counter.display_type)
+        e_reset_to = args.last('resetto', counter.get_reset_to())
+        e_reset_by = args.last('resetby', counter.reset_by)
+        e_title = args.last('title', counter.title)
+        e_desc = args.last('desc', counter.desc)
+        # Clamp the values to None if we want to remove them instead
+        _reset = e_reset if e_reset not in ('none', 'None') else None
```
zhudotexe, line 618:
```suggestion
        _reset = e_reset if e_reset.lower() != 'none' else None
```
etc

cogs5e/gametrack.py
```diff
@@ -584,6 +584,60 @@ async def customcounter_create(self, ctx, name, *args):
...
+        except InvalidArgument as e:
+            return await ctx.send(f"Failed to edit counter: {e}")
+        else:
+            value = new_counter.set(counter.value)
+            character.consumables.insert(character.consumables.index(counter), new_counter)
+            character.consumables.remove(counter)
+            await character.commit(ctx)
+
+            clamp_message = f" Clamped: {counter.value} to {value}." if not counter.value == value else ""
+            await ctx.send(f"Custom counter {name} edited."+clamp_message)
```
zhudotexe, line 639: Maybe have this display the new counter stringification (and if the value was changed, the overflow similar to `!cc <name> +/-X`)

aliasing/api/character.py
```diff
@@ -163,6 +163,79 @@ def create_cc(self, name: str, *args, **kwargs):
             self.delete_cc(name)
         return self.create_cc_nx(name, *args, **kwargs)
 
+    def edit_cc(self, name: str, minVal: str = None, maxVal: str = None, reset: str = None,
+                dispType: str = None, reset_to: str = None, reset_by: str = None,
+                title: str = None, desc: str = None):
+        """
+        Edits an existing custom counter.
+
+        Pass `none` to remove an argument entirely.
```
zhudotexe, line 172: The API should probably take `None` instead of `'none'`

aliasing/api/character.py
```diff
@@ -163,6 +163,79 @@ def create_cc(self, name: str, *args, **kwargs):
...
+        :param str minVal: The minimum value of the counter. Supports :ref:`cvar-table` parsing.
+        :param str maxVal: The maximum value of the counter. Supports :ref:`cvar-table` parsing.
+        :param str reset: One of ``'short'``, ``'long'``, ``'hp'``, ``'none'``, or ``None``.
+        :param str dispType: Either ``None`` or ``'bubble'``.
+        :param str reset_to: The value the counter should reset to. Supports :ref:`cvar-table` parsing.
+        :param str reset_by: How much the counter should change by on a reset. Supports dice but not cvars.
+        :param str title: The title of the counter.
+        :param str desc: The description of the counter.
+        :rtype: AliasCustomCounter
+        :returns: The edited counter.
```
zhudotexe, line 185: If the counter does not exist, returns `None`

aliasing/api/character.py
```diff
@@ -163,6 +163,79 @@ def create_cc(self, name: str, *args, **kwargs):
...
+        """
+        counter = self._get_consumable(name)
+
+        if minVal is not None:
+            minVal = str(minVal) if str(minVal) not in ('none', 'None') else None
+        else:
+            minVal = counter.get_min()
```
zhudotexe, line 192:
```suggestion
        if minVal is not _UNSET:
            minVal = str(minVal) if minVal is not None else None
        else:
            minVal = counter.min
```
To support using `None` to remove a value, set the default kwarg to a special `_UNSET` sentinel value (you can define this somewhere at the top of the file as `_UNSET = object()`). It might be worth pulling this logic out into a helper function (e.g. `optional_arg_or_default(argument: Optional[T], arg_t: Type[T] = str, default: Any = None, unset_sentinel: Any = _UNSET)`).

You probably also want to use `counter.min` instead of `counter.get_min()`; the latter returns the computed value, while the former returns the expression (which is probably what we want if we want to leave it unchanged).

cogs5e/gametrack.py
```diff
@@ -584,6 +584,60 @@ async def customcounter_create(self, ctx, name, *args):
...
+        `-resetto <value>` - The value to reset the counter to. Default - maximum.
+        `-resetby <value>` - Rather than resetting to a certain value, modify the counter by this much per reset. Supports dice.
+        """  # noqa: E501
+        character: Character = await ctx.get_character()
+        counter = await character.select_consumable(ctx, name)
+        # pull in the new values, or existing ones
+        args = argparse(args)
+        e_reset = args.last('reset', counter.reset_on)
+        e_max = args.last('max', counter.get_max())
+        e_min = args.last('min', counter.get_min())
```
zhudotexe, line 611: You probably want to use `counter.max` and `counter.min` - see above

cogs5e/gametrack.py
```diff
@@ -584,6 +584,60 @@ async def customcounter_create(self, ctx, name, *args):
...
+        e_type = args.last('type', counter.display_type)
+        e_reset_to = args.last('resetto', counter.get_reset_to())
```
zhudotexe, line 613: You probably want `counter.reset_to` - see above

aliasing/api/character.py
```diff
@@ -163,6 +165,49 @@ def create_cc(self, name: str, *args, **kwargs):
             self.delete_cc(name)
         return self.create_cc_nx(name, *args, **kwargs)
 
+    def edit_cc(self, name: str, minVal: str = None, maxVal: str = None, reset: str = None,
+                dispType: str = None, reset_to: str = None, reset_by: str = None,
+                title: str = None, desc: str = None):
```
zhudotexe, line 170: to make the `UNSET` sentinel work, these defaults should be `UNSET` instead of `None`

aliasing/api/character.py
```diff
@@ -163,6 +165,49 @@ def create_cc(self, name: str, *args, **kwargs):
...
+        minVal = optional_cast_arg_or_default(minVal, default=counter.min)
+        maxVal = optional_cast_arg_or_default(maxVal, default=counter.max)
+        reset = optional_cast_arg_or_default(reset, default=counter.reset)
+        dispType = optional_cast_arg_or_default(dispType, default=counter.display_type)
+        reset_to = optional_cast_arg_or_default(reset_to, default=counter.reset_to)
+        reset_by = optional_cast_arg_or_default(reset_by, default=counter.reset_by)
+        title = optional_cast_arg_or_default(title, default=counter.title)
+        desc = optional_cast_arg_or_default(desc, default=counter.desc)
+
+        if self.cc_exists(name):
```
zhudotexe, line 201: If we are raising `ConsumableException` before we get here, this is redundant (and we can remove a level of indentation!)

cogs5e/gametrack.py
```diff
@@ -584,6 +584,60 @@ async def customcounter_create(self, ctx, name, *args):
...
+        except InvalidArgument as e:
+            return await ctx.send(f"Failed to edit counter: {e}")
+        else:
+            new_value = new_counter.set(counter.value)
+            character.consumables.insert(character.consumables.index(counter), new_counter)
+            character.consumables.remove(counter)
+            await character.commit(ctx)
+
+            clamp = f"\nClamped: {new_value} ({new_value - counter.value} overflow)." if not counter.value == new_value else ""
+            await ctx.send(f"Custom counter {name} edited."+clamp)
```
zhudotexe, line 639: Even if the output wasn't clamped, I'd still output the stringification of the new counter here, to offer an at-a-glance view of whether or not the user made the right edits

zhudotexe:
Putting these magic words here for GitHub: resolves #1715 

Also, feel comfortable writing a few tests? 😛

aliasing/api/character.py
storytellermahkasad: Oh duh.

cogs5e/gametrack.py
```diff
@@ -582,7 +582,65 @@ async def customcounter_create(self, ctx, name, *args):
...
+    async def customcounter_edit(self, ctx, name, *args):
+        """
+        Edits an existing custom counter replacing passed arguments.
+
+        Pass `none` to remove an argument entirely.
+        Will clamp counter value to new limits if needed.
+
+        __Valid Arguments__
+        `-title <title>` - Sets the title for the output when modifying the counter.
+                           `[name]` will be replaced with the player's name.
```
storytellermahkasad, line 597: Indenting these doesn't work right on mobile.

PR #1721, opened 2022-02-28
Adding Coin Handling

SeanStoves:
### Summary
Adding currency tracking to Avrae.   As of this PR, functionality will be:

- DNDBeyond
-- Import currency from character sheet
-- Update DDB Sheet on changes to Avrae Coins
- Dicecloud
-- Import currency from character sheet
- GSheets
-- Import Currency from both versions v1 and v2
- Draconic Support
-- ``AliasCharacter.coinpurse`` -> ``AliasCoinpurse``
-- ``str(AliasCoinpurse)`` -> ``str``
-- ``AliasCoinpurse.pp`` -> ``int``
-- ``AliasCoinpurse.gp`` -> ``int``
-- ``AliasCoinpurse.ep`` -> ``int``
-- ``AliasCoinpurse.sp`` -> ``int``
-- ``AliasCoinpurse.ep`` -> ``int``
-- ``AliasCoinpurse[coin_type]`` -> ``int``
-- ``AliasCoinpurse.coin_str(cointype: str)`` -> ``str``
-- ``AliasCoinpurse.compact_str()`` -> ``str``
-- ``AliasCoinpurse.modify_coins(pp: int = 0, gp: int = 0, ep: int = 0, sp: int = 0, cp: int = 0)`` -- -> ``None``
-- ``AliasCoinpurse.set_coins(pp: int, gp: int, ep: int, sp: int, cp: int)`` -> ``None``
-- ``AliasCoinpurse.get_coins()`` -> ``dict``
-- ``AliasCoinpurse.parse(args)`` -> ``dict``
-- ``AliasCoinpurse.autoconvert()`` -> ``None``

### DndBeyond Functionality Show
https://user-images.githubusercontent.com/6910352/156826405-bebc9013-60ff-4da5-8cbc-6a3834c2d6e2.mp4

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/aliasing/api/character.py
+++ b/aliasing/api/character.py
@@ -688,9 +688,10 @@ class AliasCoinpurse:
         :param int sp: Silver Pieces
         :param int cp: Copper Pieces
         """
-        self._coinpurse.update_currency(pp, gp, ep, sp, cp)
+        self._coinpurse.set_currency(self._coinpurse.pp+pp, self._coinpurse.gp+gp, self._coinpurse.ep+ep,
+                                     self._coinpurse.sp+sp, self._coinpurse.cp+cp)
 
-    def set_coins(self, pp: int = 0, gp: int = 0, ep: int = 0, sp: int = 0, cp: int = 0):
+    def set_coins(self, pp: int, gp: int, ep: int, sp: int, cp: int):
         """
         Sets your coinpurse to the provided values.
 
@@ -717,6 +718,7 @@ class AliasCoinpurse:
         If the user input is a decimal number, assumes gold pieces.
         Otherwise, allows the user to specify currencies in the form '+1gp -2sp 3cp'
 
-        :rtype dict
+        :rtype tuple(dict, bool)
         """
-        return asdict(parse_coin_args(args))
+        coin_args, explicit = parse_coin_args(args)
+        return asdict(coin_args), explicit
--- a/aliasing/api/character.py
+++ b/aliasing/api/character.py
@@ -4,7 +4,6 @@ import cogs5e.models.sheet.player as player_api
 from aliasing import helpers
 from aliasing.api.statblock import AliasStatBlock
 from cogs5e.models.errors import ConsumableException
-from cogs5e.models.sheet.coinpurse import CoinsArgs
 from cogs5e.utils.gameutils import parse_coin_args
 from utils.constants import COIN_TYPES
 
@@ -673,8 +672,8 @@ class AliasCoinpurse:
         Returns a string representation of the chosen coin type.
 
         :param str cointype: The type of coin to return. ``"pp"``, ``"gp"``, ``"ep"``, ``"sp"``, and ``"cp"``
-        :return The string representation of the chosen coin type.
-        :rtype str
+        :return: The string representation of the chosen coin type.
+        :rtype: str
         """
         if cointype not in COIN_TYPES:
             raise ValueError(f"{cointype} is not valid coin.")
@@ -684,8 +683,8 @@ class AliasCoinpurse:
         """
         Returns a string representation of the compacted coin value.
 
-        :return The string representation of the compacted coin value.
-        :rtype str
+        :return: The string representation of the compacted coin value.
+        :rtype: str
         """
         return self._coinpurse.compact_string()
 
@@ -725,8 +724,8 @@ class AliasCoinpurse:
         """
         Returns a dict of your current coinpurse.
 
-        :return A dict of your current coinpurse, e.g. ``{"pp":0, "gp":1, "ep":0, "sp":-2, "cp":3}``
-        :rtype dict
+        :return: A dict of your current coinpurse, e.g. ``{"pp":0, "gp":1, "ep":0, "sp":-2, "cp":3}``
+        :rtype: dict
         """
         return self._coinpurse.to_dict()
 
@@ -737,8 +736,8 @@ class AliasCoinpurse:
         If the user input is a decimal number, assumes gold pieces.
         Otherwise, allows the user to specify currencies in the form '+1gp -2sp 3cp'
 
-        :return A dict of the coin changes, e.g. ``{"pp":0, "gp":1, "ep":0, "sp":-2, "cp":3, "total": 0.83}``
-        :rtype dict
+        :return: A dict of the coin changes, e.g. ``{"pp":0, "gp":1, "ep":0, "sp":-2, "cp":3, "total": 0.83}``
+        :rtype: dict
         """
         coin_args = parse_coin_args(args)
         return {"pp": coin_args.pp, "gp": coin_args.gp, "ep": coin_args.ep, "sp": coin_args.sp, "cp": coin_args.cp,
--- a/aliasing/api/character.py
+++ b/aliasing/api/character.py
@@ -6,6 +6,7 @@ from aliasing import helpers
 from aliasing.api.statblock import AliasStatBlock
 from cogs5e.models.errors import ConsumableException
 from cogs5e.utils.gameutils import parse_coin_args
+from utils.constants import COIN_TYPES
 
 
 class AliasCharacter(AliasStatBlock):
@@ -655,9 +656,9 @@ class AliasCoinpurse:
         self._parent_statblock = parent_statblock
 
     def __getattr__(self, item):
-        if item.lower() not in ("cp", "sp", "ep", "gp", "pp"):
+        if item not in COIN_TYPES:
             raise ValueError(f"{item} is not valid coin.")
-        return getattr(self._coinpurse, item.lower())
+        return getattr(self._coinpurse, item)
 
     def __getitem__(self, item):
         return self.__getattr__(item)
@@ -675,9 +676,9 @@ class AliasCoinpurse:
         :return The string representation of the chosen coin type.
         :rtype str
         """
-        if cointype.lower() not in ("compact", "cp", "sp", "ep", "gp", "pp"):
+        if cointype not in ("compact", "cp", "sp", "ep", "gp", "pp"):
             raise ValueError(f"{cointype} is not valid coin.")
-        return self._coinpurse.str_styled(cointype.lower())
+        return self._coinpurse.str_styled(cointype)
 
     def modify_coins(self, pp: int = 0, gp: int = 0, ep: int = 0, sp: int = 0, cp: int = 0):
         """
--- a/aliasing/api/character.py
+++ b/aliasing/api/character.py
@@ -664,22 +664,29 @@ class AliasCoinpurse:
 
     def __str__(self):
         if self._parent_statblock.options.compact_coins:
-            return self._coinpurse.coin_string('compact')
+            return self._coinpurse.compact_string()
         return str(self._coinpurse)
 
     def coin_str(self, cointype: str):
         """
         Returns a string representation of the chosen coin type.
 
-        :param str cointype: The type of coin to return. ``"compact"``, ``"pp"``, ``"gp"``, ``"ep"``, ``"sp"``, and ``"cp"``
+        :param str cointype: The type of coin to return. ``"pp"``, ``"gp"``, ``"ep"``, ``"sp"``, and ``"cp"``
         :return The string representation of the chosen coin type.
         :rtype str
         """
-        if cointype == 'compact':
-            return self._coinpurse.compact_string()
-        elif cointype not in COIN_TYPES:
+        if cointype not in COIN_TYPES:
             raise ValueError(f"{cointype} is not valid coin.")
         return self._coinpurse.coin_string(cointype)
+
+    def compact_str(self):
+        """
+        Returns a string representation of the compacted coin value.
+
+        :return The string representation of the compacted coin value.
+        :rtype str
+        """
+        return self._coinpurse.compact_string()
 
     def modify_coins(self, pp: int = 0, gp: int = 0, ep: int = 0, sp: int = 0, cp: int = 0):
         """
--- a/aliasing/api/character.py
+++ b/aliasing/api/character.py
@@ -663,18 +663,20 @@ class AliasCoinpurse:
         return self.__getattr__(item)
 
     def __str__(self):
-        return self._coinpurse.compact_str() if self._parent_statblock.options.compact_coins else str(self._coinpurse)
+        if self._parent_statblock.options.compact_coins:
+            return self._coinpurse.str_styled('compact')
+        return str(self._coinpurse)
 
     def coin_str(self, cointype: str):
         """
-        Returns a string representation of the chosen coin type
+        Returns a string representation of the chosen coin type.
 
         :param str cointype: The type of coin to return
         :rtype str
         """
-        if cointype.lower() not in ("cp", "sp", "ep", "gp", "pp"):
+        if cointype.lower() not in ("compact", "cp", "sp", "ep", "gp", "pp"):
             raise ValueError(f"{cointype} is not valid coin.")
-        return f"{self._coinpurse.to_dict().get(cointype.lower()):,} {cointype.lower()}"
+        return self._coinpurse.str_styled(cointype.lower())
 
     def modify_coins(self, pp: int = 0, gp: int = 0, ep: int = 0, sp: int = 0, cp: int = 0):
         """
--- a/aliasing/api/character.py
+++ b/aliasing/api/character.py
@@ -4,7 +4,7 @@ import cogs5e.models.sheet.player as player_api
 from aliasing import helpers
 from aliasing.api.statblock import AliasStatBlock
 from cogs5e.models.errors import ConsumableException
-from cogs5e.utils.gameutils import parse_coin_args
+from cogs5e.utils.gameutils import parse_coin_args, CoinsArgs
 from utils.constants import COIN_TYPES
 
 
@@ -712,6 +712,13 @@ class AliasCoinpurse:
         :param int cp: Copper Pieces
         """
         self._coinpurse.set_currency(pp, gp, ep, sp, cp)
+
+    def autoconvert(self):
+        """
+        Converts all of your coins into the lowest amount of coins possible.
+        100cp turns into 1gp, 5sp turns into 1ep, etc.
+        """
+        self._coinpurse.update_currency(self._coinpurse.auto_convert_up(CoinsArgs))
 
     def get_coins(self):
         """
--- a/aliasing/api/character.py
+++ b/aliasing/api/character.py
@@ -729,9 +729,9 @@ class AliasCoinpurse:
         If the user input is a decimal number, assumes gold pieces.
         Otherwise, allows the user to specify currencies in the form '+1gp -2sp 3cp'
 
-        :return A dict of the coin changes, e.g. ``{"pp":0, "gp":1, "ep":0, "sp":-2, "cp":3, "total_cp": 83}``
+        :return A dict of the coin changes, e.g. ``{"pp":0, "gp":1, "ep":0, "sp":-2, "cp":3, "total": 0.83}``
         :rtype dict
         """
         coin_args = parse_coin_args(args)
         return {"pp": coin_args.pp, "gp": coin_args.gp, "ep": coin_args.ep, "sp": coin_args.sp, "cp": coin_args.cp,
-                "total_cp": coin_args.total_cp}
+                "total": coin_args.total}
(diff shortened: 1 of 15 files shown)


cogs5e/models/sheet/coinpurse.py
```diff
@@ -0,0 +1,180 @@
...
+            self.pp + pp >= 0,
+            self.gp + gp >= 0,
+            self.ep + ep >= 0,
+            self.sp + sp >= 0,
+            self.cp + cp >= 0
+        )):
+            if explicit and not await confirm(ctx, "You don't have enough of the chosen coins to complete this transaction. "
+                                                   "Auto convert from larger coins? (Reply with yes/no)"):
+                raise InvalidArgument("You cannot put a currency into negative numbers.")
+            pp, gp, ep, sp, cp = self.auto_convert(pp, gp, ep, sp, cp)
```
zhudotexe, line 145: @Croebh 
IMO the `update_currency` method should be a primitive (i.e. synchronous, no side effects like confirming and autoconverting); rather than asking the user for confirmation and doing the conversion using internal variables here, I think we should pull that logic out:
1. Modify `CoinsArgs` to include a `strict` or `explicit` attribute (replacing the tuple that `parse_coins_args` returns)
2. Create a new method to resolve a strict CoinsArgs to a non-strict CoinsArgs that is a valid transaction on a given Coinpurse
3. Confirm with the user there

cogs5e/models/sheet/coinpurse.py
```diff
@@ -0,0 +1,180 @@
...
+
+        self.pp += pp
+        self.gp += gp
+        self.ep += ep
+        self.sp += sp
+        self.cp += cp
+
+        if self._live_integration:
+            self._live_integration.sync_coins()
```
zhudotexe, line 154: To avoid code duplication, once the autoconverting side effect is refactored out, we can replace this body with a call to `set_currency`:
```suggestion
        self.set_currency(self.pp + pp, self.gp + gp, self.ep + ep, self.sp + sp, self.cp + cp)
```
This means we don't have to have all the validation and live sync code written out twice, since `set_currency` already does all this.

cogs5e/sheets/dicecloud.py
```diff
@@ -129,6 +133,27 @@ def get_stats(self) -> BaseStats:
         self.stats = stats
         return stats
 
+    def get_coinpurse(self):
+        """ Due to the format of the coin handling in DiceCloud, this is going to be slightly messy."""
+        if self.character_data is None: raise Exception('You must call get_character() first.')
+        if self.coinpurse:
+                return self.coinpurse
+        character = self.character_data
```
zhudotexe, line 141: Unused local here

cogs5e/sheets/dicecloud.py
```diff
@@ -129,6 +133,27 @@ def get_stats(self) -> BaseStats:
...
+        coins = {
+            "pp": 0,
+            "gp": 0,
+            "ep": 0,
+            "sp": 0,
+            "cp": 0
+        }
+
+        for i in self.character_data.get('items'):
+            if re.search(r"^((plat|gold|electrum|silver|copper)( piece)?|(pp|gp|ep|sp|cp))", i['name'], re.IGNORECASE):
```
zhudotexe, line 152:
```suggestion
            if re.fullmatch(r"(platinum|gold|electrum|silver|copper)( pieces?)?|(pp|gp|ep|sp|cp)", i['name'], re.IGNORECASE):
```

cogs5e/sheets/gsheet.py
```diff
@@ -265,6 +266,9 @@ def _gchar(self):
         elif vcell:
             self.additional = TempCharacter(doc.worksheet('Additional'))
             self.version = (2, 1) if "2.1" in vcell else (2, 0) if "2" in vcell else (1, 0)
+            if self.version >= (2, 0):
+                self.inventory = TempCharacter(doc.worksheet('Inventory'))
```
zhudotexe, line 270: Was this added in GSheet v2.0? We should make sure this is not a backward-incompatible change.
cc @Croebh @thecoppersanctum

cogs5e/utils/gameutils.py
```diff
@@ -16,3 +32,84 @@ async def send_hp_result(ctx, caster, delta=None):
...
+        else:
+            cp_display_embed.description = str(character.coinpurse) + "\n"
+    else:
+        cp_display_embed.title = f"{character.name}'s {CoinTypes[coin]['name']} pieces."
+        cp_display_embed.description=f"{CoinTypes[coin]['icon']} {coin}: {character.coinpurse.to_dict()[coin]:,}"
+
+    await ctx.send(embed=cp_display_embed)
+
+
+def parse_coin_args(args: str) -> tuple[CoinsArgs, bool]:
```
zhudotexe, line 59: @Croebh 
```suggestion
def parse_coin_args(args: str) -> Tuple[CoinsArgs, bool]:
```
`tuple[T, ...]` was only added in Python 3.9; we need to use `typing.Tuple`

zhudotexe: This comment is irrelevant anyways if we refactor `CoinsArgs` to include whether it was a strict currency or not; see comment above

cogs5e/utils/gameutils.py
```diff
@@ -16,3 +32,84 @@ async def send_hp_result(ctx, caster, delta=None):
...
+
+
+async def send_current_coin(ctx, character, coin="All"):
+    """
+    Sends the current contents of the CoinPurse
+    """
+    cp_display_embed = EmbedWithColor()
+    cp_display_embed.colour = 15844367
+    cp_display_embed.set_footer(text=f"For help managing your coins, use !game coinpurse")
+    cp_display_embed.set_thumbnail(url="https://www.dndbeyond.com/attachments/thumbnails/3/929/650/358/scag01-04.png")
```
zhudotexe: @SeanStoves 
For any character-related embed, we should use `EmbedWithCharacter` - this will automatically use the character's configured color and add their portrait as a thumbnail
```suggestion
    cp_display_embed = EmbedWithCharacter(character)
```

cogs5e/utils/gameutils.py
```diff
@@ -16,3 +32,84 @@ async def send_hp_result(ctx, caster, delta=None):
...
+    cp_display_embed.add_field(name="Total Value in Purse", value=f"{CoinTypes['gp']['icon']} gp: {character.coinpurse.total}", inline=False)
```
zhudotexe:
```suggestion
    cp_display_embed.add_field(name="Total Value", value=f"{character.coinpurse.total} gp", inline=False)
```
imo since this isn't the actual currency, we shouldn't show the coin itself

cogs5e/utils/gameutils.py
```diff
@@ -16,3 +32,84 @@ async def send_hp_result(ctx, caster, delta=None):
         await caster.message_controller(ctx, f"{caster.name}'s HP: {caster.hp_str(True)}{deltaend}")
     else:
         await ctx.send(f"{caster.name}: {caster.hp_str()}{deltaend}")
+
+
+async def send_current_coin(ctx, character, coin="All"):
+    """
+    Sends the current contents of the CoinPurse
+    """
```
zhudotexe, line 40:
```suggestion
async def send_current_coin(ctx, character, coin=None):
    """
    Sends the current contents of the CoinPurse.
    If ``coin`` is passed, it must be a valid coin type and will only show the amount of that specific coin.
    """
    if coin is not None and coin.lower() not in CoinTypes:
        raise ValueError(f"{coin!r} is not a valid coin type.")
```
We should have a little bit more validation here, and `None` as a default value is a little bit cleaner than `"All"`

cogs5e/utils/gameutils.py
```diff
@@ -16,3 +32,84 @@ async def send_hp_result(ctx, caster, delta=None):
...
+    cp_display_embed = EmbedWithColor()
+    cp_display_embed.colour = 15844367
+    cp_display_embed.set_footer(text=f"For help managing your coins, use !game coinpurse")
+    cp_display_embed.set_thumbnail(url="https://www.dndbeyond.com/attachments/thumbnails/3/929/650/358/scag01-04.png")
+    cp_display_embed.add_field(name="Total Value in Purse", value=f"{CoinTypes['gp']['icon']} gp: {character.coinpurse.total}", inline=False)
+    if coin == "All":
```
zhudotexe, line 48:
```suggestion
    if coin is None:
```

cogs5e/utils/gameutils.py
```diff
@@ -16,3 +32,84 @@ async def send_hp_result(ctx, caster, delta=None):
...
+        cp_display_embed.title = f"{character.name}'s Coinpurse"
+        if character.options.compact_coins:
+            cp_display_embed.description = character.coinpurse.str_styled('compact')
+        else:
+            cp_display_embed.description = str(character.coinpurse) + "\n"
```
zhudotexe, line 51:
```suggestion
            cp_display_embed.description = str(character.coinpurse)
```

cogs5e/models/sheet/coinpurse.py
```diff
@@ -0,0 +1,180 @@
...
+    def str_styled(self, coin_type):
+        if coin_type not in ("compact", "pp", "gp", "ep", "sp", "cp"):
+            raise ValueError("coin_type must be in ('compact', 'pp', 'gp', 'ep', 'sp', 'cp')")
+
+        if coin_type == 'compact':
+            coin_value = self.total
+            coin_type = 'gp'
+        else:
+            coin_value = getattr(self, coin_type)
+        return f"{CoinTypes[coin_type]['icon']} {coin_type}: {coin_value:,}"
```
zhudotexe: Rather than `coin_type` including the compact form, we should split this into two methods - `coin_str` and `compact_str`. It's then up to the caller to use the right one based on character options.

zhudotexe: nit: I prefer `f"{CoinTypes[coin_type]['icon']} {coin_value:,} {coin_type}"` for the format of the coins; having the icon and the currency name right next to each other feels a little weird

cogs5e/utils/gameutils.py
```diff
@@ -16,3 +32,84 @@ async def send_hp_result(ctx, caster, delta=None):
...
+    else:
+        cp_display_embed.title = f"{character.name}'s {CoinTypes[coin]['name']} pieces."
```
zhudotexe, line 53:
```suggestion
        cp_display_embed.title = f"{character.name}'s {CoinTypes[coin]['name']} Pieces"
```

cogs5e/utils/gameutils.py
```diff
@@ -16,3 +32,84 @@ async def send_hp_result(ctx, caster, delta=None):
...
+        cp_display_embed.description=f"{CoinTypes[coin]['icon']} {coin}: {character.coinpurse.to_dict()[coin]:,}"
```
zhudotexe, line 54: This should probably use `Coinpurse.str_styled` (or `Coinpurse.coin_str` if we refactor it)

cogs5e/gametrack.py
```diff
@@ -175,6 +176,27 @@ async def game_shortrest(self, ctx, *args):
...
+
+        if args is None:
+            return await gameutils.send_current_coin(ctx, character)
+
+        if re.fullmatch(r'([pgesc]p)', args, re.IGNORECASE):
+            return await gameutils.send_current_coin(ctx, character, args)
+
+        coins, explicit = gameutils.parse_coin_args(args)
+        await character.coinpurse.update_currency(pp=coins.pp, gp=coins.gp, ep=coins.ep, sp=coins.sp, cp=coins.cp,
+                                                  explicit=explicit, ctx=ctx)
```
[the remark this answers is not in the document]
zhudotexe: Another thing: can we display the exact delta somewhere when a user changes their coins?

cogs5e/gametrack.py
```diff
@@ -175,6 +176,27 @@ async def game_shortrest(self, ctx, *args):
         -h - Hides the character summary output."""
         await self._rest(ctx, 'short', *args)
 
+    @game.command(name='coinpurse', aliases=['coins', 'coin'])
+    async def game_coinpurse(self, ctx, *, args = None):
+        """
+        game coinpurse (pp|gp|ep|sp|cp) <amount>
+        game coins 2gp -5sp
+        TODO: game coins <amount> ( default gold )   gp.sp.cp == parse float to convert to proper format
```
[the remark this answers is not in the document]
SeanStoves: I like how that looks, and it seems to be nicely cleaned up.

cogs5e/models/sheet/coinpurse.py
SeanStoves: I was going for the field symmetry.   The coin textual representation in a straight line.  I like straight lines.  But, I do see that perhaps moving them around might be beneficial.   I can make the change and we can see how we feel when it's in use :)

zhudotexe: They'll still be in a straight line 😄
Screenshot 2022-03-04 at 12 50 40

cogs5e/utils/gameutils.py
SeanStoves: Good point, this was originally written before we had the str_styled :). I haven't refactored the 1 currency type .  Doing that now.

aliasing/api/character.py
```diff
@@ -625,3 +639,86 @@ def __str__(self):
...
+    def __init__(self, coinpurse, parent_statblock):
+        """
+        :type coinpurse: cogs5e.models.sheet.coinpurse.Coinpurse
+        :type parent_statblock: cogs5e.models.character.Character
+        """
+        self._coinpurse = coinpurse
+        self._parent_statblock = parent_statblock
+
+    def __getattr__(self, item):
+        if item.lower() not in ("cp", "sp", "ep", "gp", "pp"):
```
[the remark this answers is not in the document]
SeanStoves: @Croebh You can also make use of the Cointypes array at the top of coinpurse.py for this I think ?

cogs5e/models/ddbsync.py
```diff
@@ -28,6 +28,19 @@ async def _do_sync_hp(self):
             character_id=int(self.character.upstream_id)
         )
 
+    async def _do_sync_coins(self):
```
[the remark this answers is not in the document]
SeanStoves: We CAN update it... It's just.. tricky.  Due to the way the items work.  We would have to pull it down, then rewrite it back up after removing the old.

cogs5e/models/sheet/coinpurse.py
SeanStoves: Commited in #7f738f372452554b93bcaa7fd8a895bcb33fdcbd

cogs5e/sheets/dicecloud.py
SeanStoves: Due to the fact that you can create your own, and the wording on dice cloud, I opted to use plat and piece over these.

If someone created a platinum pieces, I can see them setting it as plat.

Also, the item by default is called gold pieces.  However, in the item add screen it asks for name + plural name, so I wanted to be able to pickup singular name.


@zhudotexe

zhudotexe: We can use `r"(plat(inum)?|gold|electrum|silver|copper)( pieces?)?|(pp|gp|ep|sp|cp)"` to match both `plat` and `platinum`. The `s?` handles 0 or 1 `s`, so this will handle both the singular and plural case.

cogs5e/sheets/gsheet.py
```diff
@@ -381,8 +387,20 @@ def get_stats(self):
...
+        for c_type in COIN_TYPES:
+            if self.version >= (2, 1):
+                coins[c_type] = int(self.inventory.value(COIN_TYPES[c_type]['gSheet']['v2']) or 0)
+            else:
+                coins[c_type] = int(self.character_data.value(COIN_TYPES[c_type]['gSheet']['v14']) or 0)
+
+        return Coinpurse(pp=coins["pp"], gp=coins["gp"], ep=coins["ep"], sp=coins["sp"], cp=coins["cp"])
+
+    def get_levels(self):
+        if self.character_data is None: raise Exception('Y ou must call get_character() first.')
```
storytellermahkasad, line 403: Y ou might want to fix this.

cogs5e/sheets/gsheet.py
```diff
@@ -381,6 +387,18 @@ def get_stats(self):
...
         return stats
 
+    def get_coinpurse(self):
+        if self.character_data is None: raise Exception('You must call get_character() first.')
+        coins = {}
+
+        for c_type in COIN_TYPES:
+            if self.version >= (2, 1):
+                coins[c_type] = int(self.inventory.value(COIN_TYPES[c_type]['gSheet']['v2']) or 0)
+            else:
```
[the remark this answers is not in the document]
Croebh: My copy of a 2.0 sheet has coins in the same place as 1.4, so should be fine

cogs5e/models/sheet/coinpurse.py
```diff
@@ -67,6 +67,26 @@ def auto_convert(self, coins=None):
             raise InvalidArgument("You do not have enough coins to cover this transaction.")
         return coins
 
+    def auto_convert_up(self, coins=None):
```
zhudotexe, line 70: While these methods are logically similar, I think naming them `auto_convert_up` and `auto_convert_down` is a little bit of a foot-gun; semantically, these methods are doing pretty different things.

`auto_convert_down` is used to resolve a negative CoinsArgs delta to a valid amount of currency, while `auto_convert_up` probably shouldn't interact with CoinsArgs at all (I think this is the source of the confusion where you called this method with the bare CoinsArgs type above), and shuffle the currencies up in the `self` instance.

IMO, we should keep `auto_convert_down`'s name as is (or rename to something like `resolve_negative_delta`), and change this method to something like this: 
```py
    def consolidate_coins(self) -> CoinsArgs:  # no input CoinsArgs, returns the delta
        total_cp = self.total * 100
        new_pp = int(total_cp // 1000)
        total_cp -= coins.pp * 1000
        new_gp = int(total_cp // 100)
        total_cp -= coins.gp * 100
        new_ep = int(total_cp // 50)
        total_cp -= coins.ep * 50
        new_sp = int(total_cp // 10)
        total_cp -= coins.sp * 10
        new_cp = int(total_cp)

        delta = CoinsArgs(pp=new_pp - self.pp, gp=new_gp - self.gp, ep=new_ep - self.ep, sp=new_sp - self.sp, cp=new_cp - self.cp)
        self.set_currency(pp=new_pp, gp=new_gp, ep=new_ep, sp=new_sp, cp=new_cp)  # to ensure sync is fired

        return delta
```

The difference between this method being a primitive that mutates the data model, whereas I asked you to break `autoconvert_down` out, is that the user request for this primitive is a lot simpler. In natural conversation, it'd be something like this:

**autoconvert_down**:
- User: "I would like to remove 12 silver pieces" (`!g coins -12sp`)
- Bot: "you don't have enough silver, do you want to convert?" (resolve_currency, confirm)
- User: "yes"
- Bot: "ok, I can do this transaction by doing -2gp, +8sp instead" (autoconvert_down)
- User (implicitly): ok
- Bot: "ok, I applied the changes to your coinpurse" (update_currency, display delta)
- User: "cool, thanks"

**autoconvert_up**:
- User: "I would like to convert all my coins up" (`!g coins convert`)
- Bot: "ok, here you go" (autoconvert_up)
- Bot: "here's what changed:" (return and display delta)

... I don't know if that really makes sense, but my brain decided that it wanted to write a screenplay between a hypothetical user and a bot today, so here you go 😛

Croebh: Part of the issue I was having was the circular import for CoinsArgs when I tried to import it into coinpurse.py, thats why I had it sending a blank input CoinsArgs (I was going to ask you for a better solution today :p)

The rest of this makes a lot of sense, and I appreciate the screenplay!

requirements.txt
```diff
@@ -27,3 +27,6 @@ boto3==1.20.24  # used by clustering in certain scenarios
 dnspython==2.1.0  # used by mongodb
 google-auth==1.28.0
 pymongo==3.11.3  # pin this below 4.0 - TLS config changes in 4.0 are breaking
+
+# Formatting and Standardization deps
+black==22.1.0
```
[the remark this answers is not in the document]
SeanStoves: @zhudotexe Do we want to just remove this from this PR and re-add it to the new PR to enable black ?

zhudotexe (approved):
🎉

Croebh (approved):

PR #1724, opened 2022-03-08
Allow setting initial value during cc creation

Infinidoge:
### Summary
Resolves AFR-921 (not on GitHub)

Add an `initial_value` kwarg to `create_cc_nx` and the respective internal object. Also adds the `-value` flag for `!cc create`.

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/aliasing/api/character.py
+++ b/aliasing/api/character.py
@@ -156,9 +156,9 @@ class AliasCharacter(AliasStatBlock):
         :param str dispType: Either ``None`` or ``'bubble'``.
         :param str reset_to: The value the counter should reset to. Supports :ref:`cvar-table` parsing.
         :param str reset_by: How much the counter should change by on a reset. Supports dice but not cvars.
-        :param str initial_value: The initial value of the counter.
         :param str title: The title of the counter.
         :param str desc: The description of the counter.
+        :param str initial_value: The initial value of the counter.
         :rtype: AliasCustomCounter
         :returns: The newly created counter.
         """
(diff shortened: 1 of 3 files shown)


aliasing/api/character.py
```diff
@@ -109,7 +109,7 @@ def delete_cc(self, name):
         self._character.consumables.remove(to_delete)
 
     def create_cc_nx(self, name: str, minVal: str = None, maxVal: str = None, reset: str = None,
-                     dispType: str = None, reset_to: str = None, reset_by: str = None,
+                     dispType: str = None, reset_to: str = None, reset_by: str = None, initial_value: str = None,
```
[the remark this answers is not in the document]
Infinidoge: Good catch, also reordered internally to match

cogs5e/models/sheet/player.py
```diff
@@ -159,12 +159,15 @@ def new(cls, character, name, minv=None, maxv=None, reset=None, display_type=Non
...
-            initial_value = max_value
+        # set initial value if not already set
+        if initial_value is None:
+            initial_value = max(0, min_value or 0)
+            if reset_to_value is not None:
+                initial_value = reset_to_value
+            elif max_value is not None:
+                initial_value = max_value
+        else:
+            initial_value = character.evaluate_math(initial_value)
```
zhudotexe, line 170: Would this allow values outside the specified min and max as an initial value?

Infinidoge: Indeed it would, was somewhat under the assumption that a person creating the counter would create it with a value within their own ranges.

I assume it should be clamped to the minimum and maximum?

Infinidoge: How does
```py
if initial_value is None: ...
else:
    initial_value = character.evaluate_math(initial_value)
    if min_value is not None:
        initial_value = max(initial_value, min_value)
    if max_value is not None:
        initial_value = min(initial_value, max_value)
```
look?

zhudotexe: LGTM

zhudotexe (approved):

PR #1735, opened 2022-03-13
Improve gsheet import error messages

### LINKED ISSUE #1289: AFR-643 More verbose/helpful error messaging on gsheet update failure
**Feature Request:** More verbose/helpful error messaging on gsheet update failure
**Who would use it?** Whoever uses the gsheet
**What part of the bot?** gsheet import / update
**How would it work?** Maybe it could point to the row/column that it's choking on
**Why should this be added?** It seems easy to break the gsheet and very hard to fix given the cryptic and sparse error message that Avrae currently gives.

- Mobius#1442

> Upvote - Mobius#1442

> Upvote - Infinidoge#1337

> Upvote - Robbe#7964

> Upvote - purplecharmanderz#6693

Votes: +6 / -0

Croebh:
### Summary
Adds some more helpful errors to the gsheet importer/updater, to help users in identifying and fixing issues with their own sheet. Specifically, this adds the specific cell and worksheet name to the missing attribute error, and adds additional errors specifically for issues with attack parsing, coins import, and a basic validator for the portrait URL. The last one was particularly important to me, as it previously wouldn't error on import, but would cause issues when the user ran any command that would display their portrait.

Additionally, updates the message about private sheets to use the bots google service email (including locals/forks), and add a note about simply making the sheet available to anyone with a link.

Resolves #1289

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/sheets/errors.py
+++ b/cogs5e/sheets/errors.py
@@ -6,7 +6,6 @@ Created on May 6, 2017
 
 
 class MissingAttribute(Exception):
-
     def __init__(self, attribute, cell, sheet):
         self.attribute = attribute
         self.cell = cell
@@ -15,18 +14,17 @@ class MissingAttribute(Exception):
 
 
 class AttackSyntaxError(Exception):
-
     def __init__(self, attack_name, cell, sheet, error):
         self.attack_name = attack_name
         self.cell = cell
         self.sheet = sheet
         self.error = error
-        super().__init__(f"Attack syntax issue for attack '{attack_name}' in cell {cell} on sheet '{sheet}':\n"
-                         f"> {error}")
+        super().__init__(
+            f"Attack syntax issue for attack '{attack_name}' in cell {cell} on sheet '{sheet}':\n" f"> {error}"
+        )
 
 
 class InvalidImageURL(Exception):
-
     def __init__(self, sheet, error):
         self.sheet = sheet
         self.error = error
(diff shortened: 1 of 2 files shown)


cogs5e/sheets/gsheet.py
```diff
@@ -282,8 +284,16 @@ async def load_character(self, ctx, args):
...
             await self.get_character()
         except (KeyError, SpreadsheetNotFound, APIError):
+            if config.GOOGLE_SERVICE_ACCOUNT is not None:
+                email = Credentials.from_service_account_info(
+                    json.loads(config.GOOGLE_SERVICE_ACCOUNT),
+                    scopes=SCOPES).signer_email
+            else:
+                email = Credentials.from_service_account_file(
+                    "avrae-google.json",
+                    scopes=SCOPES).signer_email
```
zhudotexe: Instead of loading the credentials again on each error, I think we can use `GoogleSheet.g_client.auth.signer_email` since this `Credentials` object is stored in the client

cogs5e/sheets/errors.py
```diff
@@ -6,7 +6,35 @@
 
 
 class MissingAttribute(Exception):
```
zhudotexe, line 8: We should have these error classes inherit from `ExternalImportError` (which iirc didn't exist when I originally wrote this? 😛 )

cogs5e/sheets/gsheet.py
Croebh: Ah cool! I meant to ask about that but I was waiting till Monday haha. Tested it and it works.

PR #1737, opened 2022-03-14
feat(argparser): make add_context accept dicts

Infinidoge:
### Summary
As of currently, to add a context to a ParsedArguments object based on a dict, you would need to create a temporary ParsedArguments object in a way which was harder to read.

This has the `add_context` method internally convert the dictionary into the correct ParsedArguments object without it needing to be done on the user's side.

#### PR Type

- [ ] This PR fixes an issue.
#### Other


zhudotexe (changes requested):
I would like to see some stronger type checking on this - `dict[str, str]` is an invalid type for `ParsedArguments.from_dict()` and this is a really easy way to create a corrupt ParsedArguments state.

zhudotexe (approved):

PR #1744, opened 2022-03-27
Add customization for randchar stat rolling

Croebh:
### Summary
This PR adds a number of features to the stat rolling side of ``!randchar`` by adding server settings for:
- Dice Rolled
- Number of Sets
- Number of Stats
- Assign Stats Directly (and being able to name them)
- Minimum Total Score Required
- Maximum Total Score Required
- Required Stats Over Certain Value
- Required Stats Under Certain Value 

image
image
image

Resolves AFR-926

#### PR Type

- [ ] This PR fixes an issue.
#### Other


ui/servsettings.py
```diff
@@ -360,3 +386,322 @@ async def get_content(self):
...
+        except (ValueError, asyncio.TimeoutError):
+            await interaction.send(
+                "No valid number of sets found. Press `Set Number of Sets` to try again.", ephemeral=True
+            )
+        else:
+            await self.commit_settings()
+            await interaction.send("Your number of sets have been updated.", ephemeral=True)
+        finally:
+            button.disabled = False
+            await self.refresh_content(interaction)
```
zhudotexe, line 447: It might be worth pulling the common "get value form message" logic out into a method (I'd be happy to workshop this on Discord)

ui/servsettings.py
```diff
@@ -360,3 +386,322 @@ async def get_content(self):
...
+                    timeout=60,
+                    check=lambda msg: msg.author == interaction.author and msg.channel.id == interaction.channel_id,
+                )
+                message = input_msg.content
+                if message.lower() == "default":
+                    stats = [stat.upper() for stat in STAT_ABBREVIATIONS]
+                else:
+                    stats = message.replace(", ", ",").split(",")
+                if len(stats) != self.settings.randchar_num:
+                    print(stats, len(stats), self.settings.randchar_num)
```
zhudotexe, line 505: Use `log.debug` instead of `print` to avoid log clutter

ui/servsettings.py
```diff
@@ -360,3 +386,322 @@ async def get_content(self):
...
+                    raise ValueError
+                self.settings.randchar_stat_names = stats
+                with suppress(disnake.HTTPException):
+                    await input_msg.delete()
+            except (ValueError, asyncio.TimeoutError):
+                self.settings.randchar_straight = not self.settings.randchar_straight
```
zhudotexe, line 511:
```suggestion
                self.settings.randchar_straight = False
```
Explicit is better than implicit here

ui/servsettings.py
```diff
@@ -360,3 +386,322 @@ async def get_content(self):
...
+                check=lambda msg: msg.author == interaction.author and msg.channel.id == interaction.channel_id,
+            )
+            if ">" in input_msg.content:
+                rule_type = "gt"
+                amount, value = input_msg.content.split(">")
+                new_rule = RandcharRule(type=rule_type, amount=amount, value=value)
+            elif "<" in input_msg.content:
+                rule_type = "lt"
+                amount, value = input_msg.content.split("<")
+                new_rule = RandcharRule(type=rule_type, amount=amount, value=value)
```
zhudotexe, line 604: Does pydantic validate/cast these to int when constructing like this? I forget if it does that for `parse_obj` only or not

ui/servsettings.py
```diff
@@ -360,3 +386,322 @@ async def get_content(self):
...
+            else:
+                raise ValueError
+            if self.settings.randchar_rules is None:
+                self.settings.randchar_rules = []
```
zhudotexe, line 608: If we set the default value to `[]`, this is unnecessary

ui/servsettings.py
```diff
@@ -360,3 +386,322 @@ async def get_content(self):
...
+            self.settings.randchar_rules.append(new_rule)
```
zhudotexe, line 609: We should limit the maximum number of rules to 25, so we don't overflow the select below.

ui/servsettings.py
```diff
@@ -360,3 +386,322 @@ async def get_content(self):
...
+            name="Number of Stats",
+            value=f"**{self.settings.randchar_num}**\n"
+            f"*This is how many stat rolls it will return per set, "
+            f"allowing your players to choose between them.*",
+            inline=False,
+        )
+        embed.add_field(
+            name="Assign Stats Directly",
+            value=f"**{self.settings.randchar_straight}**\n"
+            f"**Stat Names:** {', '.join(self.settings.randchar_stat_names or [stat.upper() for stat in STAT_ABBREVIATIONS])}\n"
```
zhudotexe, line 682: This can be a function (see above)

cogs5e/charGen.py
```diff
@@ -8,15 +8,121 @@
...
+    if max_total:
+        rules.append(f"Maximum of {max_total}")
+    for m, t in over.items():
+        rules.append(f"At least {t} over {m}")
+    for m, t in under.items():
+        rules.append(f"At least {t} under {m}")
+
+    stat_rolls = []
+    # Only attempt 1000 times to achieve the desired stat sets
+    for _ in range(1000):
```
zhudotexe, line 71: I recommend using `utils.dice.PersistentRollContext` with a high limit (maybe max_rolls=5000?) in addition to this for loop; see the implementation of the `!rr` command

cogs5e/charGen.py
```diff
@@ -26,20 +132,18 @@ def __init__(self, bot):
     @commands.command(name="randchar")
     @commands.max_concurrency(1, per=commands.BucketType.user)
     async def randchar(self, ctx, level=None):
-        """Rolls up a random 5e character."""
+        """Rolls up a random 5e character.
```
zhudotexe, line 135: Let's split this into a new command; reasons copied from discord:

> - media like the DDB promo video references randchar being a certain way 
> - custom statgen rules technically aren't RAW, but a common houserule (more common on Discord imo?)
> - `!randchar` is already really overloaded with `!randchar <level>`; having a new command gives us freedom to pass statgen rules as args later down the line

cogs5e/charGen.py
zhudotexe: If you do this you'll probably want to catch the `d20.TooManyRolls` exception too

ui/servsettings.py
Croebh: Will be doing this separately from this PR, after discussion in Discord

ui/servsettings.py
Croebh: Yup!

ui/servsettings.py
Croebh: Good call

zhudotexe (approved):

PR #1746, opened 2022-03-29
Add Csetting for Coins Autoconvert

Croebh:
### Summary
This adds a csetting for how the user wants to handle autoconverting of coins if they attempt to remove more coins of explicit types than they currently have. Defaults to Always Ask, but has option for Never Convert and Always Convert.

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/ui/charsettings.py
+++ b/ui/charsettings.py
@@ -104,7 +104,8 @@ class CharacterSettingsUI(CharacterSettingsMenuBase):
             name="Cosmetic Settings",
             value=f"**Embed Color**: {color_setting_desc(self.settings.color)}\n"
             f"**Show Character Image**: {self.settings.embed_image}\n"
-            f"**Use Compact Coin Display:** {self.settings.compact_coins}",
+            f"**Use Compact Coin Display:** {self.settings.compact_coins}\n"
+            f"**Coin Conversion Mode**: {autoconvert_coins_desc(self.settings.autoconvert_coins)}",
             inline=False,
         )
         embed.add_field(
@@ -127,6 +128,13 @@ class CharacterSettingsUI(CharacterSettingsMenuBase):
                 sync_desc_lines.append(f"**Inbound Sync**: {self.settings.sync_inbound}")
             embed.add_field(name="Character Sync Settings", value="\n".join(sync_desc_lines), inline=False)
         return {"embed": embed}
+
+
+_AUTOCONVERT_SELECT_OPTIONS = [
+    disnake.SelectOption(label="Always Ask", value="0"),
+    disnake.SelectOption(label="Always Convert", value="1"),
+    disnake.SelectOption(label="Never Convert", value="-1"),
+]
 
 
 class _CosmeticSettingsUI(CharacterSettingsMenuBase):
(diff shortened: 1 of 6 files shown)


utils/settings/character.py
```diff
@@ -6,6 +6,7 @@
 
 from utils.functions import get_positivity
 from . import SettingsBaseModel
+from ..enums import CoinsAutoConvert
```
zhudotexe, line 9: nit: `from ..enums` -> `from utils.enums`; both are valid but the absolute import is preferred here

utils/enums.py
```diff
@@ -10,3 +10,9 @@ class ActivationType(enum.IntEnum):
     MINUTE = 6
     HOUR = 7
     SPECIAL = 8
+
+
+class CoinsAutoConvert(enum.IntEnum):
+    ASK = 0
+    ALWAYS = 1
+    NEVER = -1
```
zhudotexe, line 18: Why -1 instead of 2?

ui/charsettings.py
```diff
@@ -205,6 +220,13 @@ async def get_content(self):
             f"*If this is enabled, your coins will be displayed in decimal gold format.*",
             inline=False,
         )
+        embed.add_field(
+            name="Coin Conversion Mode",
+            value=f"**{autoconvert_coins_desc(self.settings.autoconvert_coins)}**\n"
+            f"*Whether or not it should ask you if you would like to auto-convert your coins if you try so spend "
+            f"more than you currently have.*",
```
zhudotexe, line 227:
```suggestion
            f"*If a coin transaction would cause you to have a negative number of a certain currency, whether to automatically convert other coins to cover the transaction.*",
```

zhudotexe (approved):

utils/settings/character.py
Croebh: Is there a way to set pycharm to do this automatically? I used the context manager to add the import

utils/enums.py
Croebh: It felt right at the time, because it was the opposite of ALWAYS, and I likened it to that of adv/dis

utils/settings/character.py
zhudotexe: Screenshot 2022-03-29 at 10 00 42
There's this but I don't think there's any more fine grained selection for absolute vs relative imports :(

Croebh: Gotcha, I'll keep an eye out for this in the future regardless!

utils/settings/character.py
```diff
@@ -176,4 +178,11 @@ def set(self, new_value):
         default="disabled",
         display_func=lambda val: "enabled" if val else "disabled",
     ),
+    "autoconvertcoins": CSetting(
+        "autoconvert_coins",
+        "number",
+        description="auto convert coins mode, 0 for ask everytime, 1 for always convert, -1 for never convert",
```
zhudotexe, line 184:
```suggestion
        description="auto convert coins mode, 0 for ask everytime, 1 for always convert, 2 for never convert",
```

PR #1747, opened 2022-03-29
Add command consume to monimage, monster, and token

storytellermahkasad:
### Summary
Now consumes the command sent for monimage, monster, and token lookup if -h is passed

#### PR Type

- [X] This PR fixes an issue.
#### Other
- [X] This PR is a breaking change (e.g. methods or parameters removed/renamed)

### DIFF
--- a/cogs5e/lookup.py
+++ b/cogs5e/lookup.py
@@ -28,7 +28,7 @@ from gamedata.lookuputils import (
 from gamedata.race import RaceFeature
 from utils import checks, img
 from utils.argparser import argparse
-from utils.functions import chunk_text, get_positivity, search_and_select, smart_trim, trim_str
+from utils.functions import chunk_text, get_positivity, search_and_select, smart_trim, trim_str, try_delete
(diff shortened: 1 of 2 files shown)


cogs5e/lookup.py
```diff
@@ -451,29 +455,33 @@ async def monimage(self, ctx, *, name: str):
...
-            if name:
-                args = (name, *args)
-            return await ctx.invoke(token_cmd, *args)
+            return await ctx.invoke(token_cmd, name)
 
         # select monster
+        token_args = argparse(name)
+        hide_name = token_args.get("h", False, bool)
+        if not name.find("-") == -1:  # remove the args from the name
+            name = name[: name.find("-")].rstrip()
```
zhudotexe, line 474: Imo we shouldn't add this logic and stick to the old `async def token(self, ctx, name=None, *args)` signature instead - this feels hacky

zhudotexe (approved):

PR #1751, opened 2022-04-08
Switch from using avatar url to the avatar key

Croebh:
### Summary
The bot was saving the entire avatar URL to the database, while the website was doing just the key. This changes it to the key in order to match how the website (which actually uses this information) handles it.

https://github.com/avrae/avrae-service/pull/48

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/utils/functions.py
+++ b/utils/functions.py
@@ -449,7 +449,7 @@ async def user_from_id(ctx, the_id):
                 "$set": {
                     "username": the_user.name,
                     "discriminator": the_user.discriminator,
-                    "avatar": the_user.display_avatar.url,
+                    "avatar": the_user.display_avatar.key,
                     "bot": the_user.bot,
                 }
             },
--- a/utils/functions.py
+++ b/utils/functions.py
@@ -449,7 +449,7 @@ async def user_from_id(ctx, the_id):
                 "$set": {
                     "username": the_user.name,
                     "discriminator": the_user.discriminator,
-                    "avatar": the_user.display_avatar.key,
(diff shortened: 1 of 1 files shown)


utils/functions.py
```diff
@@ -449,7 +449,7 @@ async def update_known_user(the_user):
                 "$set": {
                     "username": the_user.name,
                     "discriminator": the_user.discriminator,
-                    "avatar": the_user.display_avatar.url,
+                    "avatar": the_user.display_avatar.key,
```
zhudotexe, line 452: If the user has the default avatar this will save an int `[0..4]` rather than a valid avatar hash (it should save None - maybe use a check on [`the_user.avatar`](https://docs.disnake.dev/en/latest/api.html#disnake.User.avatar))

zhudotexe (commented):
nit: while this is correct and would work fine, using a ternary statement in a dict value literal is less readable; I'd suggest
```py
    async def update_known_user(the_user):
        avatar_hash = the_user.avatar.key if the_user.avatar is not None else None
        await ctx.bot.mdb.users.update_one(
            {"id": str(the_user.id)},
            {
                "$set": {
                    "username": the_user.name,
                    "discriminator": the_user.discriminator,
                    "avatar": avatar_hash,
                    "bot": the_user.bot,
                }
            },
            upsert=True,
        )
```
(which also explicitly checks for `the_user.avatar` being `None` rather than falsy - explicit is better than implicit)

zhudotexe (approved):

PR #1752, opened 2022-04-09
Expose random.choices()

ProperAttorney:
### Summary
This PR solves AFR-850 by exposing the Random function random.choices() which allows for the creation of lists of arbitrary length when given an itterable, a desired length, and either weighted odds or cumulative weighted odds

#### PR Type

- [ ] This PR fixes an issue.
#### Other
image

### DIFF
--- a/aliasing/api/functions.py
+++ b/aliasing/api/functions.py
@@ -176,6 +176,10 @@ def randchoice(seq):
     return random.choice(seq)
 
 
+def randchoices(population, weights=None, cum_weights=None, k=1):
+    return random.choices(population, weights=weights, cum_weights=cum_weights, k=k)
+
+
 # signatures
 SIG_SECRET = config.DRACONIC_SIGNATURE_SECRET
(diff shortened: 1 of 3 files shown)


zhudotexe (changes requested):
Danger! Allowing users to call `random.choices` with unbound *k* can allow users to:
- allocate huge objects (e.g. `randchoices([[0] * 10000], k=100000)`)
- lock the interpreter (e.g. `randchoices([0], k=9999999999999...)`)

as the internal implementation of `random.choices` eventually becomes a list comprehension in the form of `[population[some_index_math_here] for i in _repeat(None, k)]`.

An implementation should take into account [`DraconicInterpreter.config.max_const_len` and `.max_loops`](https://github.com/avrae/draconic/blob/master/draconic/helpers.py#L43) for *k*, and ensure that the returned list does not exceed the maximum size allowed.

zhudotexe (approved):

PR #1758, opened 2022-04-13
fix: use max(map(...)) instead of max(..., key=...) in randchoices

### LINKED ISSUE #1757: AVR-891 `randchoices` doesn't work for populations whose largest item is not comparable with an int (e.g. if your population is a list of strings)
**What is the bug?**: `randchoices` doesn't work for populations whose largest item is not comparable with an int (e.g. if your population is a list of strings)
**Severity**: High
**Steps to reproduce**: `!test {{randchoices(['thing','stuff'])}}`
**Context**: I'm pretty sure it's this line https://github.com/avrae/avrae/blob/e5f31f47553a2a9116d9dc9dc278e2d379e5677e/aliasing/api/functions.py#L183 causing the bug. `max(population, key=approx_len_of)` returns the `item` itself, not the `approx_len_of(item)`

- Lathaon#6649

Infinidoge:
Fixes #1757

### Summary
Changes the line using `max(..., key=...)` to `max(map(..., ...)`, because the former does not return the list of lengths, while the latter does.

I have not pulled this down to test it, however given how simple of a change it is, I don't expect it to run into issues.

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/aliasing/api/functions.py
+++ b/aliasing/api/functions.py
@@ -180,7 +180,7 @@ def randchoice(seq):
 def randchoices(population, weights=None, cum_weights=None, k=1):
     if k > MAX_ITER_LENGTH:
         draconic._raise_in_context(draconic.IterableTooLong, "The length of the output is too large.")
-    if max(population, key=approx_len_of) * k > MAX_ITER_LENGTH:
+    if max(map(approx_len_of, population)) * k > MAX_ITER_LENGTH:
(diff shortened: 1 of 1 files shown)


zhudotexe (approved):
LGTM but please do test it

Infinidoge:
image
Confirmed that it works properly on my test instance.

image
Also works across different objects.

PR #1760, opened 2022-04-14
Rewrite Initiative Effects, Initiative Effect Buttons

### LINKED ISSUE #955: AVR-486 Moving back in initiative and then forward again decrements effects a second time
**What is the bug?**: Moving back in initiative and then forward again decrements effects a second time.
**Severity**: Medium
**Steps to reproduce**: Have a combatant with effects. Run !init next up to their turn, run !init prev once, then run !init next again. Effect durations will be decremented again.
**Context**: Accidentally advanced combat too early, and I ran !init prev to back up a turn. When I moved forward again, I noticed the effects went down an extra turn. Relatively minor, but it seems like !init prev ought to re-increment the effect durations?

- Idran#5302
Verification: 0

zhudotexe:
### Summary
This PR completely rewrites how effects in initiative are handled from the ground up, offering the following improvements:
- Better developer experience when defining new effects (new passive effects are added by adding a classvar instead of a hacky half-parsed args system)
- Better developer experience when using passive effects (new map-reduce pattern to resolve all active effects on a combatant)
- Explicit end-round tracking so that effects do not tick twice if using `!init prev`
- Moves `Spell.cast` to `actionutils.cast_spell` to avoid importing the automation engine in gamedata
- Allows attacks granted by effects to take advantage of the full Attack model instead of simple to-hit/damage attacks
- Automatic JIT migration from old-style effects to new-style
- Lots of various typing improvements

### IEffect ButtonInteractions
- Added the ability for InitiativeEffects to specify a list of ButtonInteractions - these display whenever a combatant's status is displayed (either through `!i status` or on the start of their turn)
- Clicking a button runs an Automation script, allowing implementation of damage-over-time effects and indefinite effects (e.g. Prone) without users having to memorize commands to remove effects
- Added "Remove IEffect" Automation node to remove the triggering IEffect in a ButtonInteraction
- Added "parent" and "children" target modes to the Target Automation node to target any combatants with a parent/child effects
- Renamed the old IEffect Automation node to LegacyIEffect and added IEffect (`type: ieffect2`) to specify new models

Resolves #955 

#### PR Type

- [x] This PR fixes an issue.
#### Other


Croebh:
Biggest thing I can think of is that this breaks Draconic

PR #1774, opened 2022-05-09
Use initials for combatant names

Croebh:
### Summary
This changes how monster combatants are named when they are added to combat. There are two scenarios:

1. If the monster's name has multiple words/phrases, it will join together the first letter of each word, appending a number to make it unique: 
- ``Will-o'-Wisp`` -> ``WoW1``
- ``Monastery of the Distressed Body, Grand Master`` -> ``MotDBGM1``
 
2. If the monsters name only contains a single word, it will take the first two letters of their name, capitalize it, and then append a number to make it unique:
- ``Goblin`` -> ``GO1``
- ``Orc`` -> ``OR1``

This should help reduce ambiguity when adding groups of similar monsters together. For instance, in the old system, fighting a group of 10 Goblins and a Goblin Boss left you with ``GO1`` - ``GO11``. With this new system, it'll be ``GO1`` - ``GO10`` and a ``GB1``.

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/utils/functions.py
+++ b/utils/functions.py
@@ -528,9 +528,4 @@ def get_initials(name: str) -> str:
         return name[:2].upper()
 
     # Multiple words, find initials
-    initials = ""
-    for word in split_name:
-        if word:
-            initials += word[0]
-
-    return initials
+    return "".join(word[0] for word in split_name if word)
--- a/utils/functions.py
+++ b/utils/functions.py
@@ -511,3 +511,26 @@ def exactly_one(iterable):
     if next(iterable, sentinel) is sentinel:
         return retval
     return None
+
+
+split_regex = re.compile(r"\W+")
+
+
+def get_initials(name: str) -> str:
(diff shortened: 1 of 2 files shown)


utils/functions.py
```diff
@@ -511,3 +511,26 @@ def exactly_one(iterable):
...
+    if len(split_name) == 1:
+        return name[:2].upper()
+
+    # Multiple words, find initials
+    initials = ""
+    for word in split_name:
+        if word:
+            initials += word[0]
+
+    return initials
```
zhudotexe, line 536:
```suggestion
    return "".join(word[0] for word in split_name if word)
```
minor optimization!

zhudotexe (approved):

utils/functions.py
Croebh: Generators! Wooo

PR #1791, opened 2022-05-23
Enable Black preview

zhudotexe:
### Summary
Enable the `--preview` arg for Black CLI. This adds an additional formatting pass that ensures strings conform to the line length, splitting them across multiple lines (or merging them) where necessary. It also removes unnecessary string specifiers (like f-strings with no templates).

#### PR Type

- [ ] This PR fixes an issue.
#### Other


aliasing/helpers.py
```diff
@@ -231,7 +231,7 @@ def set_cvar(character, name, value):
     value = str(value)
     if not name.isidentifier():
         raise InvalidArgument(
-            "Cvar names must be identifiers " "(only contain a-z, A-Z, 0-9, _, and not start with a number)."
```
Croebh, line 234: What even... happened here?

zhudotexe: The non-preview run of `black` determined that the string was short enough to fit on one line, but didn't merge the two strings into a single string

Croebh (approved):

PR #1793, opened 2022-05-26
Adding Casting to Groups in Combat

SeanStoves:
### Summary
Enabling the ability to cast spells within a group in initiative

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/initiative/cog.py
+++ b/cogs5e/initiative/cog.py
@@ -1375,11 +1375,7 @@ class InitTracker(commands.Cog):
                 return await ctx.send("Combatant not found.")
 
         if isinstance(combatant, CombatantGroup):
-            choices = []  # list of (name, caster, attack)
-            for com in combatant.get_combatants():
-                choices.append((f"({com.name})", com, spell_name))
-
-            _, combatant, spell_name = await get_selection(ctx, choices, lambda choice: choice[0], message="Select the Caster.")
+            combatant = await get_selection(ctx, combatant.get_combatants(), key=lambda com: com.name, message="Select the caster.")
 
         is_character = isinstance(combatant, PlayerCombatant)
         if is_character and combatant.character_owner == str(ctx.author.id):
--- a/cogs5e/initiative/cog.py
+++ b/cogs5e/initiative/cog.py
@@ -22,7 +22,7 @@ from cogsmisc.stats import Stats
 from gamedata.lookuputils import select_monster_full, select_spell_full
 from utils import checks, constants
 from utils.argparser import argparse
-from utils.functions import confirm, get_guild_member, search_and_select, try_delete, get_initials
+from utils.functions import confirm, get_guild_member, search_and_select, try_delete, get_initials, get_selection
 from . import (
     Combat,
     CombatOptions,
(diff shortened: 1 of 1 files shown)


cogs5e/initiative/cog.py
```diff
@@ -1375,7 +1375,11 @@ async def _cast(self, ctx, combatant_name, spell_name, args):
                 return await ctx.send("Combatant not found.")
 
         if isinstance(combatant, CombatantGroup):
-            return await ctx.send("Groups cannot cast spells.")
+            choices = []  # list of (name, caster, attack)
+            for com in combatant.get_combatants():
+                choices.append((f"({com.name})", com, spell_name))
+
+            _, combatant, spell_name = await get_selection(ctx, choices, lambda choice: choice[0], message="Select the Caster.")
```
zhudotexe, line 1382: Why the 3-tuple for the option list here?
```suggestion
            combatant = await get_selection(ctx, combatant.get_combatants(), key=lambda com: com.name, message="Select the caster.")
```
I believe this should be functionally equivalent.

zhudotexe: To explain why:

Let's look at the type of `get_selection()` (slightly abridged): `get_selection(ctx, choices: list[T], key: Callable[[T], str]) -> T`. Fundamentally, this function takes a list of options of any type (`list[T]`), and returns one (`T`) based on the user's selection.

So how does it display the list of options to the user if `list[T]` might not be a list of strings? That's where `key: Callable[[T], str]` comes in: it's a function that transforms each of the options into a string, and that's what gets shown to the user in the menu.

For example, let's see what happens if we run the following: 
```py
choices = ["foo", "bar", "bletch"]
await get_selection(ctx, choices, key=lambda elem: elem[::-1])  # the key is the string, reversed
```

```
Multiple Matches Found
Which one were you looking for? (Type the number or c to cancel)
[1] - oof
[2] - rab
[3] - hctelb
```

If the user selects one of these, `get_selection` returns the element that was used to render the choice in the menu - that is, `elem` rather than `key(elem)`. This means that if the user selects `2`, it would return `"bar"` rather than `"rab"`.

We extend this concept to the list of Combatants - the user sees a list of names to choose from, since our key is `lambda com: com.name`, but when the user actually selects one, `get_selection` returns the Combatant that was used to display that name.

zhudotexe (approved):
LGTM - just needs a pass with `black .` to satisfy linting

---

Copying the `get_selection` writeup here so it's not hidden within a collapsed comment:

Let's look at the type of `get_selection()` (slightly abridged): `get_selection(ctx, choices: list[T], key: Callable[[T], str]) -> T`. Fundamentally, this function takes a list of options of any type (`list[T]`), and returns one (`T`) based on the user's selection.

So how does it display the list of options to the user if `list[T]` might not be a list of strings? That's where `key: Callable[[T], str]` comes in: it's a function that transforms each of the options into a string, and that's what gets shown to the user in the menu.

For example, let's see what happens if we run the following: 
```py
choices = ["foo", "bar", "bletch"]
await get_selection(ctx, choices, key=lambda elem: elem[::-1])  # the key is the string, reversed
```

```
Multiple Matches Found
Which one were you looking for? (Type the number or c to cancel)
[1] - oof
[2] - rab
[3] - hctelb
```

If the user selects one of these, `get_selection` returns the element that was used to render the choice in the menu - that is, `elem` rather than `key(elem)`. This means that if the user selects `2`, it would return `"bar"` rather than `"rab"`.

We extend this concept to the list of Combatants - the user sees a list of names to choose from, since our key is `lambda com: com.name`, but when the user actually selects one, `get_selection` returns the Combatant that was used to display that name.

PR #1565, opened 2021-07-07
Unify help and args for madd, add, join

storytellermahkasad:
### Summary
Unifies the structure, documentation, and args for add/madd/join

<details><summary>ADD Changes</summary>

NEW -n <number or dice> 
NEW -neutral <damage type>

NO -b | Part of the command
NO -name | Part of the command
NO -rollhp | Doesn't have HP
NO -phrase | Not embed
NO -thumb | Not embed

NOTE -n command doesn't add # if only adding one (to prevent "Duke Erik1")</details>

<details><summary>JOIN Changes</summary>

NEW -hp <hp>
NEW -thp <thp>
NEW -ac <ac>
NEW -resist <damage type>
NEW -immune <damage type>
NEW -vuln <damage type>
NEW -neutral <damage type>

NO -controller | That's silly
NO -n | Also Silly
NO -name | Silly
NO -rollhp | Maybe? But likely no</details>

<details><summary>MADD Changes</summary>

NEW -controller <controller>
NEW -resist <damage type>
NEW -immune <damage type>
NEW -vuln <damage type>
NEW -neutral <damage type>

NO -phrase | Not embed
NO -thumb | Not embed</details>

Resolves #1307 
Resolves #1368 

#### PR Type

- [X] This PR is a code change that implements a feature request.
- [X] This PR fixes an issue.
- [X] This PR adds a new feature that is not an open feature request.
- [ ] This PR is not a code change (e.g. documentation, README, ...)
#### Other
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [x] If code changes were made then they have been tested.
- [X] I have updated the documentation to reflect the changes.

PR #1813, opened 2022-06-15
Add the ability for "init add" to take different ability scores and more

ProperAttorney:
### Summary
This PR allows the "init add" ability to take different values for ability scores, proficiency bonus, save proficiencies, skill proficiencies, and skill expertise.

This PR attempts to address AFR-886.

#### PR Type

- [ ] This PR fixes an issue.
#### Other


cogs5e/initiative/cog.py
```diff
@@ -227,9 +252,24 @@ async def add(self, ctx, modifier: int, name: str, *args):
             init_roll_skeleton = str(init)
 
         me = Combatant.new(
-            name, controller, init, init_skill, hp, ac, private, Resistances.from_dict(resists), ctx, combat
+            name, controller, init, init_skill, hp, ac, private, Resistances.from_dict(resists), ctx, combat, stats
         )
 
+        for skill in profs:
+            skill_key = await search_and_select(ctx, SKILL_NAMES, skill, lambda s: s)
```
zhudotexe, line 259: Since AFAIK the intended user for this interface is programmatic, I would make this a strict check that the given arg was a valid skill key instead (including some documentation).

See the implemention of the Ability Check automation node or `cogs5e.initiative.effects.passive#resolve_check_advs` for examples of this.

cogs5e/initiative/cog.py
```diff
@@ -227,9 +252,24 @@ async def add(self, ctx, modifier: int, name: str, *args):
...
+            me.skills[skill_key].prof = 1
+            me.skills[skill_key].value += me.stats.prof_bonus
```
zhudotexe, line 261: If I pass the same skill multiple times, will this give me superpowers?

cogs5e/initiative/cog.py
```diff
@@ -227,9 +252,24 @@ async def add(self, ctx, modifier: int, name: str, *args):
             init_roll_skeleton = str(init)
 
         me = Combatant.new(
-            name, controller, init, init_skill, hp, ac, private, Resistances.from_dict(resists), ctx, combat
+            name, controller, init, init_skill, hp, ac, private, Resistances.from_dict(resists), ctx, combat, stats
```
zhudotexe, line 255: I'd reorganize the order of these args so `ctx` and `combat` aren't kind of floating in the middle there - I can handle that after the other changes

ProperAttorney:
> Generally this seems like a pretty solid PR. I would move some of the arg handling logic out of cog.py and into its own file (let's do a new file, like `combatant_builders.py` - this also gives a home for further work on PRs like #1565) to avoid bloating the cog file.

Do you know what parts of the arg handling you would like to keep in the cog.py file?

zhudotexe:
Uhhh, whatever makes sense to you. Tbh we can probably move the whole thing.

cogs5e/initiative/cog.py
```diff
@@ -164,104 +164,53 @@ async def begin(self, ctx, *args):
         await ctx.send(out)
 
     @init.command()
-    async def add(self, ctx, modifier: int, name: str, *args):
-        """Adds a generic combatant to the initiative order.
-        Generic combatants have a 10 in every stat and +0 to every modifier.
+    async def add(self, ctx, name: str, *args):
```
zhudotexe, line 167: Is the removal of the `modifier` arg an intended change? This is technically a breaking change.

cogs5e/initiative/combatant_builders.py
```diff
@@ -0,0 +1,115 @@
...
+from utils.constants import STAT_ABBR_MAP
+from utils.argparser import argparse
+from cogs5e.initiative.effects.passive import resolve_check_advs, resolve_save_advs
+from .utils import create_combatant_id
+from . import (
+    Combatant,
+    utils,
+)
+
+async def add_builder(ctx, combat, name, args):
```
zhudotexe, line 14: IMO the builder function should stay synchronous, as the async signature implies that the function has some side effects. See comments below for some tips to make this sync

cogs5e/initiative/combatant_builders.py
```diff
@@ -0,0 +1,115 @@
...
+    args = argparse(args)
```
zhudotexe, line 15: This should probably be moved outside the builder function, and the builder function should take `args: ParsedArguments` rather than `args: str`

cogs5e/initiative/combatant_builders.py
```diff
@@ -0,0 +1,115 @@
...
+    controller = ctx.author.id
+    group = None
+    hp = None
+    ac = None
+    resists = {}
+    adv = args.adv(boolwise=True)
+
+    if combat.get_combatant(name, True) is not None:
+        await ctx.send("Combatant already exists.")
+        return
```
zhudotexe, line 27: This should be moved outside the builder - the builder isn't responsible for checking that what it builds is legal to add, it just builds.

cogs5e/initiative/combatant_builders.py
```diff
@@ -0,0 +1,115 @@
...
+
+    id = create_combatant_id()
+
+    if args.last("controller"):
+        controller_name = args.last("controller")
+        member = await commands.MemberConverter().convert(ctx, controller_name)
+        controller = member.id if member is not None and not member.bot else controller
```
zhudotexe, line 34: It's really annoying that this is async, ugh.
Ok, we might need to leave the entire function async just to use this, since it has nontrivial logic to look up a member if it's not in cache.

Alternatively, we could move this out and pass the controller as an arg, but that feels a little weird.

cogs5e/initiative/combatant_builders.py
```diff
@@ -0,0 +1,115 @@
...
+    for skill in exps:
+        skills[skill].prof = 2
+        skills[skill].value += 2*stats.prof_bonus
+
+    if args.get("p"):
+        init = args.last("p",type_=int)
+    else:
+        init = roll(skills.initiative.d20()).total
+
+    if args.adv:
```
zhudotexe, line 69: args.adv() is a method

cogs5e/initiative/combatant_builders.py
```diff
@@ -0,0 +1,115 @@
...
+        skills.initiative.adv = adv
```
zhudotexe, line 70: This logic seems backwards - setting the adv after the roll means that the adv is not considered in the roll.

Also what about the `modifier` command arg?

cogs5e/initiative/combatant_builders.py
```diff
@@ -0,0 +1,115 @@
...
+    for k in ("resist", "immune", "vuln"):
+        resists[k] = args.get(k)
+
+    if args.last("ac"):
+        ac = args.last("ac", type_=int)
+
+    if args.last("hp"):
+        hp = args.last("hp", type_=int)
+        if hp < 1:
+            return await ctx.send("You must pass in a positive, nonzero HP with the -hp tag.")
```
zhudotexe, line 88: this should be `raise InvalidArgument` instead of `return await ctx.send` to ensure the command ends if this is invalid

cogs5e/initiative/combatant_builders.py
```diff
@@ -0,0 +1,115 @@
...
+        skills = skills,
+        saves = saves,
+        resistances = Resistances.from_dict(resists),
+        ac = ac,
+        max_hp = hp,
+        temp_hp = thp,
+        creature_type = creature_type,
+    )
+
+    return me
```
zhudotexe, line 115: Consider how to return the initiative roll result as well, rather than just displaying the final result.

zhudotexe (approved):

zhudotexe:
/ci-approve

let's see if this works

cogs5e/initiative/combatant_builders.py
```diff
@@ -0,0 +1,134 @@
...
+        saves[save_key].value += stats.prof_bonus
+
+    # resistances
+    resists = {}
+    for k in ("resist", "immune", "vuln"):
+        resists[k] = args.get(k)
+    resistances = Resistances.from_dict(resists)
+
+    # ac
+    ac = None
```
Croebh, line 95: Couldn't this simply be `ac = args.last("ac", type_=int)`, since if the arg isnt present it resolves to None?

Croebh (approved):

PR #1900, opened 2023-01-10
Jan 9th Hotfix

Croebh:
### Summary
Discord's API updated earlier tonight to have stricter type enforcement on Falsey values, which broke a number of base commands and a significant amount of user aliases. This updates Avrae to the newest 2.7.0 version of Disnake, which supports that API change. 

In addition to changes required to handle those Falsey values, the library update also changed the `AutoShardedBot` class to not use `**kwargs`, and so this also updates the code to use specific kwargs in the `super().__init__()` and not just star unpacking all of them.

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/models/embeds.py
+++ b/cogs5e/models/embeds.py
@@ -44,7 +44,7 @@ class EmbedWithCharacter(disnake.Embed):
         super().__init__(**kwargs)
         if name:
             self.set_author(name=character.name)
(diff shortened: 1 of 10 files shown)


Infinidoge (approved):
Everything looks good to me.
Smoke-tested on my instance, works as intended.

PR #1934, opened 2023-06-14
June 14th Hotfix

Croebh:
### Summary
- Move DC Bonus adding to after checking if they have a DC
  - If its a spell using the default DC, and the user is not a spellcaster, this was causing an error before it would check of the DC was None

Plan to go live with this PR is at 10pm MT, June 14th

#### PR Type

- [x] This PR fixes an issue.
#### Other


zhudotexe (approved):
oops

PR #1943, opened 2023-07-10
Updating Black to latest

SeanStoves:
### Summary

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/sheets/dicecloudv2.py
+++ b/cogs5e/sheets/dicecloudv2.py
@@ -706,9 +706,7 @@ class DicecloudV2Parser(SheetLoaderABC):
         verb, proper = (
             ("casts", True)
             if atk_prop["type"] == "spell"
-            else ("uses", False)
-            if atk_prop["actionType"] != "attack"
-            else (None, False)
+            else ("uses", False) if atk_prop["actionType"] != "attack" else (None, False)
         )
         log.debug(f"Parsing {atk_prop['type']}")
         activation = ACTIVATION_MAP.get(atk_prop["actionType"], ActivationType.SPECIAL)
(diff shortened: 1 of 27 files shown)


cogs5e/sheets/dicecloudv2.py
```diff
@@ -706,9 +706,7 @@ def parse_attack(self, atk_prop) -> Attack:
         verb, proper = (
             ("casts", True)
             if atk_prop["type"] == "spell"
-            else ("uses", False)
-            if atk_prop["actionType"] != "attack"
-            else (None, False)
+            else ("uses", False) if atk_prop["actionType"] != "attack" else (None, False)
```
tlake, line 709: this change _looks_ pretty bizarre to me, but it might not affect anything

tests/e2e/cogs/dice_test.py
```diff
@@ -118,11 +118,11 @@ async def test_inline_rolling_reaction(avrae, dhttp, mock_ldclient):
     async with server_settings(avrae, inline_enabled=InlineRollingType.REACTION):
         with mock_ldclient.flags({"cog.dice.inline_rolling.enabled": True}):
             avrae.message("[[1d20]]")
-            await dhttp.receive_reaction("\N{game die}")
+            await dhttp.receive_reaction("\N{GAME DIE}")
```
tlake, line 121: is this `game die` change (and the one that follows on line 125) actually case-sensitive?

cogs5e/sheets/dicecloudv2.py
SeanStoves: I'm sadly, not finding any documentation on a configuration flag we can use to disable this particular check with Black.

tlake: i guess we can either leave it and let black do black to it, or we can try to rewrite it as a more pythonic if/else block?

i think this would do the same thing:
```
verb, proper = None, False
if atk_prop["type"] == "spell":
    verb, proper = "casts", True
elif atk_prop["actionType"] != "attack":
    verb, proper = "uses", False
```

it's not important anyway, just a formatting preference where black and i differ xD

PR #1948, opened 2023-09-06
Adding an extra redis Set command to RedisIO

SeanStoves:
### Summary
Resolving items without expiration not being created in redis, This was caused by redis-py update performed awhile ago.   By doing a basic set if we don't need the other values, we will now properly save status' such as inline reaction messages for the onboarding process ( first time messages )

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/dice/inline.py
+++ b/cogs5e/dice/inline.py
@@ -159,7 +159,7 @@ class InlineRoller:
             await user.send(embed=embed)
         except disnake.HTTPException:
             return
-        await self.bot.rdb.set(f"cog.dice.inline_rolling.users.{user.id}.onboarded.message", str(time.time()))
+        await self.bot.rdb.bset(f"cog.dice.inline_rolling.users.{user.id}.onboarded.message", str(time.time()))
 
     async def inline_rolling_reaction_onboarding(self, user):
         if await self.bot.rdb.get(f"cog.dice.inline_rolling.users.{user.id}.onboarded.reaction"):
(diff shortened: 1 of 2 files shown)


zhudotexe:
It looks like the error was introduced by https://github.com/avrae/avrae/commit/3c3e45e4bc61d43228310b7ea7eb407594bedf0b#diff-ff0d6746a7959a58415318a9c2b04d1bf87b371b5f2f05b69ac6b325373752ddR29, which changed the default behaviour of `.set` to have an expiration time of 1 second rather than never - why not just revert this? The change seems unintentional

SeanStoves:
> It looks like the error was introduced by [3c3e45e#diff-ff0d6746a7959a58415318a9c2b04d1bf87b371b5f2f05b69ac6b325373752ddR29](https://github.com/avrae/avrae/commit/3c3e45e4bc61d43228310b7ea7eb407594bedf0b#diff-ff0d6746a7959a58415318a9c2b04d1bf87b371b5f2f05b69ac6b325373752ddR29), which changed the default behaviour of `.set` to have an expiration time of 1 second rather than never - why not just revert this? The change seems unintentional

IIRC, it failed w/ an expiration of 0.  1+ if there's an ex

zhudotexe:
Looks like `ex` expects `None` rather than `0` in redis-py. https://redis-py.readthedocs.io/en/stable/commands.html#redis.commands.core.CoreCommands.set

PR #1953, opened 2023-12-01
Uprev disnake to fix issue with display_name

ianfhunter:
### Summary
Currently avrae cannot retrieve server display names since discord changed its display name setup. [2.9.0 disnake apparently fixes this in 2.9.0](https://docs.disnake.dev/en/stable/whats_new.html#v2-9-0), so this PR upgrades to that version. 

Current behaviour: 
`!test {{verify_signature(signature()).author.display_name}}` falls back to the same result as `!test {{verify_signature(signature()).author.name}}` (username).

This change should allow server usernames to be accessible via the `verify_signature` function once more.

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ d20==1.1.2
 aiobotocore==2.1.0
 redis==4.6.0
 cachetools==4.2.2
-disnake~=2.7.0
+disnake~=2.9.0
 gspread==3.7.0
 httplib2==0.19.0
 launchdarkly-server-sdk==7.2.0


ianfhunter:
Would the GH actions be enough evidence? They appear to run all the pytests including end-to-end tests. I'm not sure what you would be looking for further than that

ianfhunter:
I can't get all the unit tests to run locally on this branch *or* the current main. I'm not sure why, but i suppose it is some delta with my setup

 - tests/e2e/cogs/dice_test.py::test_inline_rolling_reaction
 - tests/e2e/cogs/dice_test.py::test_inline_rolling_enabled

The two above both fail for me on both branches, so indicate nothing. Again, I would appreciate it if we could run the github actions, which is a known good environment.

The rest of the tests have passed with this change.

1drturtle:
Ran the tests on my local copy and all passed - another thing is might want to double check wherever we use the `display_name` field in the project and ensure it's the correct new one.

ianfhunter:
I had a look at usages of `display_name` and didn't see anything problematic. Some references are maybe non-specific like the name used in echoing, but I think it's all as it was before

PR #1958, opened 2024-01-03
Add support for Bestiary Builder bestiaries & Fix critterDB regexes

VeryGreatFrog:
### Summary
This Pull Request does two things:
It adds support for bestiaries from https://bestiarybuilder.com/, a bestiary website designed for use with Avrae and convenience. 
It also fixes some very minor issues in the critterDB bestiary regex parsers to match current statblock formatting.
#### PR Type
- [ ] This PR fixes an issue.
#### Other
- [X] This PR is a breaking change (e.g. methods or parameters removed/renamed)

### Notes
This pull request changes the bestiary schema to add a new field. I tested this change and I believe it is fine, but for my piece of mind I would appreciate if someone else also tested them.

### DIFF
--- a/cogs5e/homebrew.py
+++ b/cogs5e/homebrew.py
@@ -85,11 +85,10 @@ class Homebrew(commands.Cog):
 
     @bestiary.command(name="import")
     async def bestiary_import(self, ctx, url):
-        """
-        Imports a bestiary from [Bestiary Builder](https://bestiary.stevnbak.dk) or [CritterDB](https://critterdb.com/).
+        """Imports a bestiary from [Bestiary Builder](https://bestiary.stevnbak.dk) or [CritterDB](https://critterdb.com/).
         **Bestiary Builder:**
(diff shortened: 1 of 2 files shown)


cogs5e/homebrew.py
```diff
@@ -85,9 +85,11 @@ async def bestiary_delete(self, ctx, *, name):
 
     @bestiary.command(name="import")
     async def bestiary_import(self, ctx, url):
-        """
-        Imports a bestiary from [CritterDB](https://critterdb.com/).
+        """Imports a bestiary from [Bestiary Builder](https://bestiary.stevnbak.dk) or [CritterDB](https://critterdb.com/).
```
DemiplaneJoseph, line 88: Should this be bestiarybuilder.com, like the code below?

VeryGreatFrog: Good catch, fixed!

PR #1978, opened 2024-01-18
assigning asyncio lock instance to bot object for concurrency control

RishiReddywofc:
### Summary
Changing the self object to bot object for the garbage collection method to handle the memory leaks caused by disnake version upgrade

### Changelog Entry

#### PR Type

- [X] This PR fixes an issue.
#### Other

### DIFF
--- a/dbot.py
+++ b/dbot.py
@@ -120,7 +120,7 @@ class Avrae(commands.AutoShardedBot):
         self.glclient.init()
 
         # lock for garbage collection
-        self.gc_lock = asyncio.Lock()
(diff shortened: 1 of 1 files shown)


dbot.py
```diff
@@ -120,7 +120,7 @@ def __init__(self, prefix, description=None, **options):
         self.glclient.init()
 
         # lock for garbage collection
-        self.gc_lock = asyncio.Lock()
+        bot.gc_lock = asyncio.Lock()
```
SeanStoves, line 123: This should have stayed as self.gc_lock

PR #1980, opened 2024-01-19
Enable image import from bestiary builder

VeryGreatFrog:
This enables image imports from bestiary builder, which previously were accidentally left empty.

### DIFF
--- a/cogs5e/models/homebrew/bestiary.py
+++ b/cogs5e/models/homebrew/bestiary.py
@@ -415,7 +415,7 @@ def _monster_factory_bestiary_builder(data, bestiary_name):
         la_per_round=data["la_per_round"],
         attacks=attacks,
         proper=data["proper"],
-        image_url="",
+        image_url=data["image_url"],
         spellcasting=spellcasting,
         homebrew=True,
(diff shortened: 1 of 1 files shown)


cogs5e/models/homebrew/bestiary.py
```diff
@@ -415,7 +415,7 @@ def _monster_factory_bestiary_builder(data, bestiary_name):
         la_per_round=data["la_per_round"],
         attacks=attacks,
         proper=data["proper"],
-        image_url="",
+        image_url=data["image_url"],
```
SeanStoves, line 418: @VeryGreatFrog If there's no image, does your export give a blank value for this, or are we going to have KeyErrors ?

VeryGreatFrog: It will be an empty string: https://github.com/Bestiary-Builder/Bestiary-Builder/blob/main/backend/logic/bestiaries.ts#L556

SeanStoves: Excellent, let's get this merged into nightly for testing then :)

PR #2000, opened 2024-02-27
Add support for !character channel and !character server reset_all commands

etriebe:
### Summary
Implementing feature request: Add an option to support per channel for characters in addition to per sheet
char-command

### Changelog Entry
Add two new parameters to the !character command: `!char channel` and `!char server reset_all`. The first will 

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/models/character.py
+++ b/cogs5e/models/character.py
@@ -19,7 +19,7 @@ from cogs5e.models.sheet.spellcasting import Spellbook, SpellbookSpell
 from cogs5e.models.sheet.statblock import DESERIALIZE_MAP as _DESER, StatBlock
 from cogs5e.models.sheet.coinpurse import Coinpurse
 from cogs5e.sheets.abc import SHEET_VERSION
-from utils.functions import search_and_select
+from utils.functions import confirm, search_and_select
 from utils.settings import CharacterSettings
 
 log = logging.getLogger(__name__)
@@ -156,10 +156,13 @@ class Character(StatBlock):
         return cls(**d)
 
     @classmethod
-    async def from_ctx(cls, ctx, ignore_guild: bool = False):
+    async def from_ctx(cls, ctx, ignore_guild: bool = False, ignore_channel: bool = False):
         owner_id = str(ctx.author.id)
         active_character = None
-        if ctx.guild is not None and not ignore_guild:
+        if ctx.channel is not None and not ignore_channel:
+            channel_id = str(ctx.channel.id)
+            active_character = await ctx.bot.mdb.characters.find_one({"owner": owner_id, "active_guilds": channel_id})
+        if ctx.guild is not None and not ignore_guild and active_character is None:
             guild_id = str(ctx.guild.id)
             active_character = await ctx.bot.mdb.characters.find_one({"owner": owner_id, "active_guilds": guild_id})
         if active_character is None:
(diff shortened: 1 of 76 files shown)


cogs5e/sheetManager.py
```diff
@@ -436,30 +436,77 @@ async def character(self, ctx, *, name: str = None):
 
     @character.command(name="server")
     @commands.guild_only()
-    async def character_server(self, ctx):
+    async def character_server(self, ctx, args=""):
         """
         Sets the current global active character as a server character.
         If the character is already the server character, unsets the server character.
 
+        args:
```
LazyDope, line 444: Since the function of reset_all is pretty separate from the function of normal `!char server`, I would say it would be better to make it a subcommand, rather than use the args here.

etriebe: Fair enough! Will make documentation much easier to understand. 

Do you think we would reset all and undo the server being set too?

LazyDope: It makes the most sense, and there's no reason not to.

etriebe:
/home/avrae/cogs5e/sheetManager.py:516: RuntimeWarning: coroutine 'Character.unset_server_active' was never awaited
  server_character.unset_server_active(ctx)

etriebe:
@LazyDope can you take a look again? Been a few weeks and I haven't heard back from anyone.

cogs5e/sheetManager.py
```diff
@@ -488,13 +488,14 @@ async def character_channel(self, ctx):
 
         await try_delete(ctx.message)
 
-    @character.command(name="reset_all")
+    @character.command(name="resetall")
     @commands.guild_only()
-    async def reset_all(self, ctx):
+    async def resetall(self, ctx):
```
LazyDope, line 493: Use the snake case name for the function, only change the command name with the decorator (as you have done)

cogs5e/sheetManager.py
```diff
@@ -488,13 +488,14 @@ async def character_channel(self, ctx):
...
         """
         This will unset any channel or server-specific characters that have been set and force the current global character to be used everywhere on this server
         """  # noqa: E501
 
+        listOfUnsetCharacters = []
```
LazyDope, line 498: snake case variable names please

cogs5e/models/character.py
```diff
@@ -365,11 +409,11 @@ async def set_server_active(self, ctx):
             raise NoPrivateMessage()
         guild_id = str(ctx.guild.id)
         owner_id = str(ctx.author.id)
-        # unset anyone else that might be active on this server
+        # unset anyone else that might be active with this server/channel id
```
LazyDope, line 412: These should no longer say /channel since they only apply to guild

cogs5e/sheetManager.py
```diff
@@ -436,31 +436,92 @@ async def character(self, ctx, *, name: str = None):
 
     @character.command(name="server")
     @commands.guild_only()
-    async def character_server(self, ctx):
+    async def character_server(self, ctx, args=""):
```
LazyDope, line 439: Don't need these args

cogs5e/models/character.py
```diff
@@ -156,15 +160,18 @@ def from_dict(cls, d):
         return cls(**d)
 
     @classmethod
-    async def from_ctx(cls, ctx, ignore_guild: bool = False):
+    async def from_ctx(cls, ctx, ignore_guild: bool = False, ignore_channel: bool = False, raise_error: bool = True):
```
LazyDope, line 163: Should use a try-except statement instead of disabling the error, the rest of the function is not safe to run if active_character is None

Croebh:
As a note, they're not the 'arbiter' of if this gets merged, @SeanStoves is. @LazyDope is just a very helpful and active contributor

LazyDope:
I am happy to take another look, but just because I'm out of notes by no means makes it complete.

SeanStoves:
Yeah, @Croebh and @LazyDope are amazing people.   I haven't had the time to review this due to other concerns that have been occurring, and it being a rather core functionality add I want to thoroughly look over it, or if one of the other more active members approve it, I'll move it over to nightly for testing.

LazyDope:
I will mention that my recent review doesn't cover everything, it was just the first things I noticed.

etriebe:
Great! Thanks for taking a look so far @LazyDope!

cogs5e/models/character.py
etriebe: Fixed, though it isn't as graceful as I wanted since I'll need each individual call to be try-except'ed, because each call might fail with NoCharacter and I wouldn't want the subsequent calls for the next character to not pass.

etriebe:
OK, all review feedback so far addressed and 'black <filename>' ran on both files to address workflow failures.

PR #2005, opened 2024-03-10
fix(character): return existing cc in create_cc_nx

Infinidoge:
### Summary
If `create_cc_nx` doesn't create a CC, it will return the CC that already exists.

### Changelog Entry

#### PR Type

- [ ] This PR fixes an issue.
#### Other


Infinidoge:
Lint failure isn't even from me >:(

PR #2017, opened 2024-03-26
Add a method to allow adding to cache from sheetManager

etriebe:
### Summary
The cache wasn't being populated when a character was being added by name selection, so it'd make a cache error be thrown.

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/models/character.py
+++ b/cogs5e/models/character.py
@@ -210,6 +210,11 @@ class Character(StatBlock):
         return inst
 
     @classmethod
+    async def add_to_cache(cls, owner_id, upstream, character):
+        cls._cache[owner_id, upstream] = character
+        return
+
+    @classmethod
     async def from_bot_and_channel_id(cls, ctx, owner_id: str, channel_id: str):
         owner_id = str(owner_id)
         channel_id = str(channel_id)
--- a/cogs5e/models/character.py
+++ b/cogs5e/models/character.py
@@ -512,7 +512,7 @@ class Character(StatBlock):
         if ctx.channel is None:
             raise NoPrivateMessage()
         channel_id = str(ctx.channel.id)
-        return await self.unset_active_channel_helper(ctx, channel_id)
+        return await self.unset_active_channel_helper(ctx, channel_id, previous_character)
 
     async def unset_active_channel_helper(self, ctx, channel_id, previous_character):
         channel_id = str(channel_id)
--- a/cogs5e/models/character.py
+++ b/cogs5e/models/character.py
@@ -378,7 +378,8 @@ class Character(StatBlock):
             self._live_integration.commit_soon(ctx)  # creates a task to commit eventually
 
     async def set_active(self, ctx):
-        """Sets the character as globally active and unsets any server-active character or channel-active character in the current context, whichever is most specific."""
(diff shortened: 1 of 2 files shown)


etriebe:
@SeanStoves should we merge this in?

SeanStoves:
> @SeanStoves should we merge this in?

Still noodling this one.

cogs5e/sheetManager.py
```diff
@@ -448,7 +448,7 @@ async def get_character_by_name(self, ctx, name):
             ctx, user_characters, name, lambda e: e["name"], selectkey=lambda e: f"{e['name']} (`{e['upstream']}`)"
         )
 
-        return Character.from_dict(selected_char)
+        return Character.serialize_character_from_dict(str(ctx.author.id), char)
```
devgr, line 451: Looks like you meant `selected_char` rather than `char`

cogs5e/models/character.py
```diff
@@ -374,6 +380,8 @@ async def commit(self, ctx, do_live_integrations=True):
 
     async def set_active(self, ctx):
         """Sets the character as globally active and unsets any server-active character or channel-active character in the current context, whichever is most specific."""
+        owner_id = str(ctx.author.id)
+        did_unset_active_location = False
```
devgr, line 384: Looks like these variables don't get used in the function?

cogs5e/models/character.py
```diff
@@ -209,6 +209,12 @@ async def from_bot_and_ids(cls, bot, owner_id: str, character_id: str):
         cls._cache[owner_id, character_id] = inst
         return inst
 
+    @classmethod
+    async def serialize_character_from_dict(cls, owner_id, character_dictionary):
+        char = Character.from_dict(character_dictionary)
+        cls._cache[owner_id, character_dictionary["upstream"]] = character_dictionary
```
devgr, line 215: Based on the other uses of the cache in this file, it looks like the cache should be holding a `Character` instance, not the character dictionary. Maybe you meant `cls._cache[owner_id, character_dictionary["upstream"]] = char`?

cogs5e/models/character.py
```diff
@@ -209,6 +209,12 @@ async def from_bot_and_ids(cls, bot, owner_id: str, character_id: str):
         cls._cache[owner_id, character_id] = inst
         return inst
 
+    @classmethod
+    async def serialize_character_from_dict(cls, owner_id, character_dictionary):
```
devgr, line 213: `deserialize` might make more sense here than `serialize`, since the result is an object instance rather than a string or dictionary.

cogs5e/models/character.py
etriebe: They do on line 416 & 419

cogs5e/models/character.py
etriebe: Ah, good point

cogs5e/models/character.py
etriebe: Ah, now I see... This was due to merge. My local git I was responding to comments from didn't have the merge done yet.

devgr (approved):
Code looks good! I will leave it up to Sean on what to do next.

PR #2026, opened 2024-04-01
Avr 419

kevklash:
### Summary

### Changelog Entry

- User creation updated to use context instead of dicts within internal/utility functions
- Ldclient initialization has been updated
- Utility functions "variation_for_discord_user" and "variation_for_ddb_user" were updated to build context where needed instead of creating users out of a dict. Argument references where updated, but not the type as it is hard to determine the type/format of all args since they are not using type hints. This will avoid errors in exceptions for passing args in the wrong format/type.
- Flag evaluation remains the same, "variation" built-in method accepts the same args, but reference to the users were changed to context where needed.

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/utils/feature_flags.py
+++ b/utils/feature_flags.py
@@ -32,9 +32,9 @@ class AsyncLaunchDarklyClient(ldclient.LDClient):
         """Return a variation for a key given a DDB user or None."""
         if user is None:
             # TODO: Check if second argument is valid for "anonymous" as a -kind- of user
-            context = ldclient.Context.create(str(discord_id))
+            user = ldclient.Context.create(str(discord_id))
         else:
-            user = user.to_ld_dict()
+            user = ldclient.Context.create(str(user.to_ld_dict()))
         return await self.variation(key, user, default)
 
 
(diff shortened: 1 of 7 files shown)


SeanStoves:
Code looks good @kevklash , if you can run ``black .`` on your branch then push the commit up.   Please make sure you are using black v24.1.1.

utils/feature_flags.py
```diff
@@ -15,26 +16,29 @@ class AsyncLaunchDarklyClient(ldclient.LDClient):
...
         """Return a variation for a key given a discord user."""
-        return await self.variation(key, discord_user_to_dict(user), default)
+        return await self.variation(key, discord_user_to_context(user), default)
 
     async def variation_for_ddb_user(self, key: str, user: Optional["ddb.auth.BeyondUser"], default, discord_id: int):
         """Return a variation for a key given a DDB user or None."""
         if user is None:
-            user = {"key": str(discord_id), "anonymous": True}
+            # TODO: Check if second argument is valid for "anonymous" as a -kind- of user
+            context = ldclient.Context.create(str(discord_id))
```
SeanStoves, line 35: I think this might be a breaking change.   Should this be reverted to user vs context?  Mainly because line 38 expects user to exist and we aren't setting it if it is coming to this function as none.

SeanStoves: I'm not sure exactly how to test this though.

PR #2049, opened 2024-05-29
Improve monster embeds

Lathaon:
### Summary
Make monster embeds closer to source material.
- Alignment can be omitted
- Size/Type/Alignment line is italicized
- Trait/Action names are italicized and have a full stop instead of a colon
- Removed punctuation not present in books
- Changed "CR" to "Challenge" as it is in books
- XP is comma-separated

Note: I haven't done anything with how ability scores are displayed because Discord doesn't support markdown tables and the current solution seems good enough while that remains true.

### Changelog Entry
Make monster embeds closer to source material.

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/lookup.py
+++ b/cogs5e/lookup.py
@@ -612,19 +612,19 @@ class Lookup(commands.Cog):
         if visible:
             embed_queue[-1].description = monster.get_meta()
             if monster.traits:
-                trait = "\n\n".join(f"**{a.name}:** {a.desc}" for a in monster.traits)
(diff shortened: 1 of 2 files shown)


VeryGreatFrog:
It will need testing to see if changing CR to Challenge doesn't cause the line to wrap to a new one on mobile, which I think would be a regression

Lathaon:
> It will need testing to see if changing CR to Challenge doesn't cause the line to wrap to a new one on mobile, which I think would be a regression

I tested with a CR 30 creature and it didn't wrap for me.

PR #2059, opened 2024-07-30
HP deltas

### LINKED ISSUE #1821: AFR-960 QoL: make automation footer display health deltas
Rather than one line per hp change displaying the new HP, group up all the HP changes and display the final new HP per target along with a delta

Lathaon:
### Summary
Resolves #1821 - sorta. HP deltas are shown but damage instances are **not** grouped together, as per [benubu](https://github.com/benubu)'s comment on that FR.

NOTE: This changes the return value of `Combatant.hp_str()`, which may affect aliases.
Instead of `<10/10> (+10 temp)` for a non-hidden combatant, it will display as `<10/10, 10 temp>`.
This change was necessary to make sure the new HP is distinct from the delta HP.

### Changelog Entry
When a combatant is damaged via automation, HP deltas will now be shown in the footer alongside the new HP values.

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/initiative/combatant.py
+++ b/cogs5e/initiative/combatant.py
@@ -151,15 +151,17 @@ class Combatant(BaseCombatant, StatBlock):
         """Returns a string representation of the combatant's HP."""
         out = ""
         if not self.is_private or private:
+            hp_strs = []
             if self.max_hp is not None and self.hp is not None:
-                out = f"<{self.hp}/{self.max_hp} HP>"
+                hp_strs.append(f"{self.hp}/{self.max_hp} HP")
             elif self.hp is not None:
-                out = f"<{self.hp} HP>"
-            else:
-                out = ""
+                hp_strs.append(f"{self.hp} HP")
 
             if self.temp_hp and self.temp_hp > 0:
-                out += f" (+{self.temp_hp} temp)"
+                hp_strs.append(f"{self.temp_hp} temp")
+
+            out = f"<{', '.join(hp_strs) or 'None'}>"
+
         elif self.max_hp is not None and self.max_hp > 0:
             ratio = self.hp / self.max_hp
             if ratio >= 1:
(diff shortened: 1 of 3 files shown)


SeanStoves:
Since I'm not heavy on using or creating aliases.  What kind of impact could this change have ?

SeanStoves:
@Lathaon .  We would love to merge this, but can we get clarification of what kind of issues this can cause for existing functionality?

Lathaon:
> @Lathaon . We would love to merge this, but can we get clarification of what kind of issues this can cause for existing functionality?

It will change the output of Combatant.hp_str()

What would have looked like `<10/10> (+5 temp)` will look like `<10/10 HP, 5 temp>`

I can't think of any obvious problems that would arise from the return value just being displayed in an embed or echo.

If an alias is doing further processing on the hp_str instead of the hp and temp_hp values directly then that might be affected.

If an alias is manually constructing a string to resemble the output of hp_str then it would no longer match.

I don't know which, if any, aliases would fall into these categories.

SeanStoves:
> > @Lathaon . We would love to merge this, but can we get clarification of what kind of issues this can cause for existing functionality?
> 
> It will change the output of Combatant.hp_str()
> 
> What would have looked like `<10/10> (+5 temp)` will look like `<10/10 HP, 5 temp>`
> 
> I can't think of any obvious problems that would arise from the return value just being displayed in an embed or echo.
> 
> If an alias is doing further processing on the hp_str instead of the hp and temp_hp values directly then that might be affected.
> 
> If an alias is manually constructing a string to resemble the output of hp_str then it would no longer match.
> 
> I don't know which, if any, aliases would fall into these categories.

So if I'm understanding properly, this will only impact those aliases that people have written that parses the hp_str() in code to do different things?   I think we can add that to the changelog as a headsup and move on.  I'll get this merged in fro the next update.

1drturtle:
Seems like it is failing tests though - they also might need to be updated

SeanStoves:
> Seems like it is failing tests though - they also might need to be updated

Yep, I'll be taking a peek at the tests at some point and see why it's failing. But, with PHB 2024 releasing in just a few short weeks for preorders, I'm focusing on that.

PR #2080, opened 2024-09-08
Working on spellbook fix to handle both versions(2014 and 2024)

kevklash:
### Summary

- Fixing spellbook to handle spells for both versions: 2014 and 2024
- Csettings update: Implemented a toggle to switch between versions(2014 AND 2024). We will read these settings to handle spells according to the version configured on the character.
- On import, the argument(either 2014 or 2014) specifies the version, defaulting to 2024 if no arg is provided

#### PR Type

- [x] This PR fixes an issue.
#### Other


cogs5e/sheetManager.py
```diff
@@ -919,6 +919,11 @@ async def import_sheet(self, ctx, url: str, *, args=""):
         if character and beyond_match:
             await send_ddb_ctas(ctx, character)
 
+        # Update charsetting based on the version argument
+        character.options.version = version
+        await character.options.commit(ctx.bot.mdb, character)
```
LazyDope, line 924: Not sure how early this is, but this probably should be in `SheetManager._load_sheet` so that you don't have to commit twice.

kevklash: I think it is good to put it in ._load_sheet, but I have to commit twice though:
1. character.commit for the character model
2. character.options.commit for the character settings.

@SeanStoves Does it makes sense?

SeanStoves: > I think it is good to put it in ._load_sheet, but I have to commit twice though:
> 
> 1. character.commit for the character model
> 2. character.options.commit for the character settings.
> 
> @SeanStoves Does it makes sense?

@kevklash I do see the point of putting it in ``_load_sheet``.   If we put it in that function, it will be good across the board.  However, I'm concerned about passing the version into that. and how further uses of it will be.

LazyDope: `character.options` is included in a `Character.commit` based on `Character.to_dict`, it seems that `CharacterSettings.commit` is just a convenience method for when you don't need to commit the whole character, however in this case we need to commit the whole character anyways so might as well.

SeanStoves: > `character.options` is included in a `Character.commit` based on `Character.to_dict`. It seems that `CharacterSettings.commit` is just a convenience method for when you don't need to commit the whole character, however in this case we need to commit the whole character anyways so might as well.

Sounds good.. @kevklash Let's give that a whirl in that function.   We should be solid ;)

LazyDope: Also might I suggest using a command argument like `-version` to specify a version, rather than a positional argument? We're already passing `args` to `_load_sheet` so you don't actually need to change the function signature for either `import` (compared to master) or `_load_sheet` to make this change if done this way.

kevklash: @SeanStoves @LazyDope Noted! Making changes and testing locally now.

cogs5e/sheetManager.py
```diff
@@ -933,7 +929,7 @@ async def beyond(self, ctx, url: str, *, args=""):
         await self.import_sheet(ctx, url, args=args)
 
     @staticmethod
-    async def _load_sheet(ctx, parser, args, loading):
+    async def _load_sheet(ctx, parser, args, loading, version):
```
LazyDope, line 931: I still think getting the version from `args` rather than passing it through as a function parameter works better here, since you're already parsing the args regardless when you load the character, you might as well pull that out into its own variable and check that when you are setting the version.

PR #2109, opened 2025-01-02
AFR-1014: Add AliasRoles

1drturtle:
### Summary

Adds a list of AliasRole to AliasAuthor
Accounted for DMs

### Changelog Entry

AliasRoles can now be accessed from `ctx.author.get_roles()`

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/aliasing/api/context.py
+++ b/aliasing/api/context.py
@@ -253,12 +253,12 @@ class AliasAuthor:
         """
         return self._roles
 
-
     def __str__(self):
         return f"{self.name}#{self.discriminator}"
 
     def __repr__(self):
         return f"<{self.__class__.__name__} name={self.name!r} id={self.id!r}>"
+
 
 class AliasRole:
     """
@@ -293,6 +293,7 @@ class AliasRole:
 
     def __repr__(self):
         return f"<{self.__class__.__name__} name={self.name!r} id={self.id!r}>"
+
 
 class AliasCategory:
     """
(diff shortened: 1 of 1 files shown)


aliasing/api/context.py
```diff
@@ -206,6 +206,9 @@ def __init__(self, author):
         self._id = author.id
         self._discriminator = author.discriminator
         self._display_name = author.display_name
+        self._roles = []
+        if isinstance(author, disnake.Member):
+            self._roles = [AliasRole(role) for role in author.roles]
```
LazyDope, line 211: Maybe use a generator here so the AliasRoles are generated upon request at not at the start of every alias
```suggestion
            self._roles = (AliasRole(role) for role in author.roles)
```

aliasing/api/context.py
```diff
@@ -243,13 +246,56 @@ def display_name(self):
         """
         return self._display_name
 
+    @property
+    def roles(self):
+        """
+        The user's roles. When used in a DM, it is always empty.
+        """
+        return self._roles
```
LazyDope, line 254:
```suggestion
    def roles(self):
        """
        The user's roles. When used in a DM, it is always empty.
        """
        return list(self._roles)
```

LazyDope (changes requested):

aliasing/api/context.py
LazyDope: Ignore this, forgot generators are single use

aliasing/api/context.py
LazyDope: Even without the generator, this would probably be a beneficial change, as we don't want the list inside `ctx.author` to be modified.

kevklash (approved):
LGTM

PR #2119, opened 2025-02-21
Update conc_parent logic to check if we really want "none" as a parent

Corvux89:
### Summary
In certain cases when casting spell automations, if an effect is applied to a target with no parent set, it will default to the conc_parent. This is giving user's the option to set the parent node to "none", and have the conc_parent be ignored. 

### Changelog Entry
Allow for "none" to be set in the ieffect parent 

#### PR Type

- [X] This PR fixes an issue.
#### Other


codybanman (changes requested):
Looks great! Could you rebase the branch onto nightly and update the ieffect section in docs/automation_ref.rst to describe the new behavior?

PR #2131, opened 2025-03-04
Read servsettings in Draconic

Lathaon:
### Summary
Implements AFR-1069 (Allow read-only access of servsettings in Draconic)

### Changelog Entry
Adds `ctx.guild.servsettings()`

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/aliasing/evaluators.py
+++ b/aliasing/evaluators.py
@@ -171,7 +171,6 @@ class ScriptingEvaluator(draconic.DraconicInterpreter):
         uvars = await helpers.get_uvars(ctx)
         inst = cls(ctx, builtins=DEFAULT_BUILTINS, initial_names=uvars)
         inst._cache["uvars"].update(uvars)
-        # I'm just putting this comment here to see if the tests actually run this time
         if ctx.guild:
             if hasattr(ctx, "get_server_settings"):
                 inst._cache["servsettings"] = await ctx.get_server_settings()


aliasing/evaluators.py
```diff
@@ -170,6 +170,11 @@ async def new(cls, ctx):
         uvars = await helpers.get_uvars(ctx)
         inst = cls(ctx, builtins=DEFAULT_BUILTINS, initial_names=uvars)
         inst._cache["uvars"].update(uvars)
+        if ctx.guild:
+            if hasattr(ctx, "get_server_settings"):
+                inst._cache["servsettings"] = await ctx.get_server_settings()
+            else:
+                inst._cache["servsettings"] = await ServerSettings.for_guild(mdb=ctx.bot.mdb, guild_id=ctx.guild.id)
```
1drturtle, line 177: I don't see where ServerSettings is being imported, is it an implicit import?

Lathaon: No, you're right. Good catch, haven't got time to sort that out now but will do it tomorrow

Lathaon: Okay that should be sorted now. Still working fine in my local copy. Although I don't think any of the slash commands use Draconic so the `else` might not actually be reachable in the first place...

PR #2162, opened 2025-03-29
Improve Local Test Docs

1drturtle:
### Summary

- Cleans up a few warnings in the tests
- Adds documentation for running tests on docker (including `docker compose down` beforehand to prevent intermittent dice_test failures.
- removes version tag from docker compose (deprecated, shows a warning when it is present)
- adds `r` to some raw-format strings to reduce test warnings
- gate Publicity and Stat loops from running without a token or production env, respectively
- **Change redis close mechanic** from .close to .aclose (showed as deprecated in test warnings) (only important change to review)

### Changelog Entry
none (not user facing)

#### PR Type

- [X] This PR fixes an issue.
#### Other

### DIFF
--- a/utils/argparser.py
+++ b/utils/argparser.py
@@ -154,7 +154,7 @@ class ParsedArguments:
     # Have to escape the _ in the type_ parameter for docs
     # noinspection PyIncorrectDocstring
     def get(self, arg, default=None, type_=str, ephem=False):
-        """
+        r"""
         Gets a list of all values of an argument.
 
         :param str arg: The name of the arg to get.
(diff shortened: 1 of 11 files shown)


utils/argparser.py
```diff
@@ -154,7 +154,7 @@ def empty_args(cls):
     # Have to escape the _ in the type_ parameter for docs
     # noinspection PyIncorrectDocstring
     def get(self, arg, default=None, type_=str, ephem=False):
-        """
+        r"""
```
Croebh, line 157: Does this break readthedocs?

1drturtle: not when I built it locally!

PR #2163, opened 2025-04-01
Avr 544 gather bi data

kevklash:
### Summary
Add a stream client to the Avrae Bot to stream BI data

#### PR Type

- [ ] This PR fixes an issue.


dbot.py
```diff
@@ -361,6 +364,36 @@ async def on_command(ctx):
...
+        # send command to kafka
+        producer = KafkaProducer(KafkaProducer.config)
+        avrae_command = {
+            "EVENT_TIME": time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime()),
+            "PLATFORM": "discord",
+            "MESSAGE_ID": ctx.message.id,
+            "MESSAGE_NAME": "AVRAE_COMMAND",
+            "DISCORD_ID": ctx.message.author.id,
+            "DDB_USER_ID": ctx.message.author.name,
+            "DISCORD_SERVER_ID": ctx.message.guild.id,
```
1drturtle, line 380: not sure if this is intended, but this will not produce an event in case of a command in DM.

dbot.py
```diff
@@ -361,6 +364,36 @@ async def on_command(ctx):
...
+            "ACTION_DESCRIPTORS": [
+                {
+                    "COMMAND_ID": ctx.command.qualified_name,
+                    "COMMAND_CATEGORY": ctx.command.cog_name,
+                    "SUBCOMMAND_ID": None,
+                    "ARGS": ctx.message.content.split(" ")[1:],
+                }
+            ],
+        }
+        producer.produce(
```
1drturtle, line 392: Will this block the async thread?

dbot.py
```diff
@@ -361,6 +364,36 @@ async def on_command(ctx):
             "cmd: chan {0.message.channel} ({0.message.channel.id}), serv {0.message.guild} ({0.message.guild.id}), "
             "auth {0.message.author} ({0.message.author.id}): {0.message.content}".format(ctx)
         )
+        print(
```
1drturtle, line 367: can we leave this as a debug?

requirements.txt
```diff
@@ -29,3 +29,6 @@ pymongo==4.7.2
 
 # Handling for Datadog
 ddtrace~=2.21.3
+
+# BI Data Collection
+confluent-kafka
```
SeanStoves, line 34: @kevklash , Let's pin this at minimum to a major version so it doesn't break in the future.

utils/config.py
```diff
@@ -76,3 +76,11 @@
...
 # ---- discord bot list ----
 DBL_TOKEN = os.getenv("DBL_TOKEN")  # optional
+
+# ---- kafka ----
+KAFKA_BOOTSTRAP_SERVER = os.getenv("KAFKA_BOOTSTRAP_SERVER")
+KAFKA_API_KEY = os.getenv("KAFKA_API_KEY")
+KAFKA_API_SECRET = os.getenv("KAFKA_API_SECRET")
+KAFKA_SECURITY_PROTOCOL = os.getenv("KAFKA_SECURITY_PROTOCOL", "SASL_SSL")
+KAFKA_SASL_MECHANISM = os.getenv("KAFKA_SASL_MECHANISM", "PLAIN")
+KAFKA_ACKS = os.getenv("KAFKA_ACKS", "all")
```
SeanStoves, line 86: @kevklash , Let's set some default values here, so that it doesn't break for our community developers.

confluent_client/producer.py
```diff
@@ -0,0 +1,47 @@
...
+        'bootstrap.servers': config.KAFKA_BOOTSTRAP_SERVER,
+        'sasl.username': config.KAFKA_API_KEY,
+        'sasl.password': config.KAFKA_API_SECRET,
+        'security.protocol': config.KAFKA_SECURITY_PROTOCOL,
+        'sasl.mechanisms': config.KAFKA_SASL_MECHANISM,
+        'acks': config.KAFKA_ACKS
+    }
+    
+    def __init__(self, config):
+        self.producer = Producer(config)
```
SeanStoves, line 27: @kevklash , Let's make this Feature Flag enabled.  That way it doesn't try to run this code when someone does local development.   For an example of this, take a look at the gridfs code, where it uses grid if it see's the LD flag, else it just loads the local json.

SeanStoves (changes requested):

dbot.py
SeanStoves: @kevklash , Yeah, as Turtle states, you will want to take the ctx.message.guild.id and give it a default value, or accept None in Kafka's side.

dbot.py
SeanStoves: @kevklash , That's a good question, will this run in the background w/o breaking anything if a connection takes to long?

PR #2174, opened 2025-05-29
revert(AVR-914): revert changes in v4.2.63 back to v4.2.62

codybanman:
### Summary
Something cause the bot to stop responding to requests that went out in v4.2.63. The changes themselves appear benign and they don't appear to cause issues on nightly. We are going to revert the changes and then release them one by one to see which change caused the problem..

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/aliasing/api/context.py
+++ b/aliasing/api/context.py
@@ -94,7 +94,7 @@ class AliasGuild:
     Represents the Discord guild (server) an alias was invoked in.
     """
 
-    def __init__(self, guild, servsettings=None):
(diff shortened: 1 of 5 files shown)


Croebh:
Did this reversion not already happen last week when the bot was breaking? Or did y'all do it all via AWS and not in GitHub like we normally do-

codybanman:
@Croebh Correct, the actual reversion happened via AWS last week. This PR is syncing the repo with the current deployed state.

PR #2210, opened 2025-08-28
feat: update SimpleCombatant save and check to include ieffects and csetttings

7he4lph4:
### Summary
Adds combat init effects and character settings bonus modifiers to SimpleCombatant.save() and .check() methods (.check() is new).

### Changelog Entry
Modifiers from combat init effects and character settings are now added to SimpleCombatant save() and check() methods (.check() is new).

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/aliasing/api/combat.py
+++ b/aliasing/api/combat.py
@@ -9,6 +9,7 @@ from cogs5e.models.errors import InvalidSaveType
 from cogs5e.models.automation.utils import parse_save_bonuses
 from cogs5e.models.sheet.statblock import StatBlock
 from utils.argparser import ParsedArguments
+from utils.constants import SKILL_MAP
 from . import validators
 
 if TYPE_CHECKING:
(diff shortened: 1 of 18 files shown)


1drturtle (approved):
overall looks great! would love some tests for the changes to save, new check function, and the progressive unique matching.

tonip-wizard:
Could you just clear the conflicts?

VeryGreatFrog:
This is a significant breaking change. Aliases that previously used .save and .check in combat, but also included their own code to detect combat bonuses would count the bonuses twice, while having no way to announce as such to the user. Older aliases might thus unexpectedly break with no real period of warning. Has it been considered how this will be addressed? Perhaps this new behaviour should be behind a consider_combat = True flag?

7he4lph4:
Only the `.save()` is breaking yes. IIRC, there is no `.check()` and this PR would be adding it. When this was discussed, the idea was to make a post in [#alias-announcement](https://discord.com/channels/269275778867396608/620791391165087754) announcing this new breaking change. But having back compatibility sounds nice, we could add in the opt-in flag, maybe instead of a `consider_combat`, we can have `include_ieffects = True` as well as a `include_csettings = True` for finer control.

7he4lph4:
This change is no longer breaking and fully backwards compatible. Those 2 new flags have been added, this can be tested with:
### Setup combat
```
!multiline
!csettings reroll 1
!csettings talent true
!init begin
!init join
!init effect <character name> TestEffect -sadv dex -cadv perception -dc 2 -cb 1d4
```
### Test checks and saves on combatant with flags
```
!test <drac2>
c = combat()
me = c.me

out = [
  "default save: " + me.save("dex").full,
  "ieffect save: " + me.save("dex", include_ieffects=True).full,
  "csettings save: " + me.save("dex", include_csettings=True).full,
  "both save: " + me.save("dex", include_ieffects=True,
  include_csettings=True).full,
  "default check: " + me.check("perception").full,
  "ieffect check: " + me.check("perception", include_ieffects=True).full,
  "csettings check: " + me.check("perception", include_csettings=True).full,
  "both check: " + me.check("perception", include_ieffects=True,
  include_csettings=True).full,
]

return "\n".join(out)
</drac2>
```

7he4lph4:
@tonip-wizard This PR is no longer a breaking change so I'd say you could remove the risky tag and possibly add it to the review queue. Thank you.

PR #2217, opened 2025-09-11
Add `-tickon` parameter in !i effect to support tick_on_combatant_id

justhalf:
### Summary
Even though `InitiativeEffect` has the `tick_on_combatant_id` argument to specify when an effect timer would tick (as seen in https://github.com/avrae/avrae/blob/nightly/cogs5e/initiative/effects/effect.py#L99), it is not yet included as one of the parameters in `!i effect` command.

This PR adds the `-tickon <combatant>` additional argument to `!i effect` to support that use case.

### Changelog Entry
Add `-tickon <combatant>` argument to `!i effect`

#### PR Type
- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/initiative/cog.py
+++ b/cogs5e/initiative/cog.py
@@ -1010,15 +1010,16 @@ class InitTracker(commands.Cog):
             else:
                 targets.append(target)
 
-        tickon = None
-
-        for t in [target_name] + args.get("tickon")[-1:]:
+        tick_on_combatant_id = None
+        tickon_arg = args.last("tickon")
+        if tickon_arg:
             tickon = await combat.select_combatant(
                 ctx,
-                t,
+                tickon_arg,
                 f"Pick the combatant in whose turn this effect timer would tick.",
                 select_group=False,
             )
+            tick_on_combatant_id = tickon.id
 
         duration = args.last("dur", -1, int)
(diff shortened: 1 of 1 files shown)


cogs5e/initiative/cog.py
```diff
@@ -1009,6 +1010,16 @@ async def effect(self, ctx, target_name: str, effect_name: str, *, args=""):
...
 
+        tickon = None
+
+        for t in [target_name] + args.get("tickon")[-1:]:
+            tickon = await combat.select_combatant(
+                ctx,
+                t,
+                f"Pick the combatant in whose turn this effect timer would tick.",
+                select_group=False,
+            )
```
codybanman: This is clever and works, but I think it could you simplify it to something like this: 
```python
tick_on_combatant_id = None
tickon_arg = args.last("tickon")
if tickon_arg:
    tickon = await combat.select_combatant(
        ctx,
        tickon_arg,
        "Pick the combatant in whose turn this effect timer would tick.",
        select_group=False,
    )
    tick_on_combatant_id = tickon.id
```
It took my Friday brain too long to understand the loop the first time I read over it 😅

justhalf: Ah, good point. That's more easily understandable, and keeps the default value of None as before.

PR #2223, opened 2025-10-26
feat: add initscore flag to !i madd/meta

7he4lph4:
### Summary
Adds `initscore` flag to `!init madd` and `!init meta`. Uses monster's initiative score (10 + modifier) instead of rolling, as per [MM2024 optional rule](https://www.dndbeyond.com/sources/dnd/mm-2024/how-to-use-a-monster#Initiative).

### Changelog Entry
Added `initscore` flag to `!init madd` and `!init meta` as per MM2024 optional rule.

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/initiative/cog.py
+++ b/cogs5e/initiative/cog.py
@@ -116,6 +116,7 @@ class InitTracker(commands.Cog):
         `dyn` - Dynamic initiative; Rerolls all initiatives at the start of a round.
         `turnnotif` - Notifies the controller of the next combatant in initiative.
         `deathdelete` - Disables deleting monsters below 0 hp.
+        `initscore` - Use initiative scores instead of rolling initiative for monsters by default.
         `-name <name>` - Sets a name for the combat instance.
         `norecord` - If the server has opted in to the Natural Language Training project, omit this combat from recording.
         """  # noqa E501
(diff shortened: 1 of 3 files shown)


cogs5e/initiative/cog.py
```diff
@@ -288,14 +295,20 @@ async def madd(self, ctx, monster_name: str, *, args=""):
...
             check_roll = None  # to make things happy
-            if p is None:
+            if p is not None:
+                init = int(p)
+            elif initscore:
+                init = 10 + init_skill.value
+                if adv is True:
+                    init += 5
+                elif adv is False:
+                    init -= 5
```
codybanman: This ignores `-b`, I think it would be good to add it here if set.

codybanman (changes requested):
This looks great. I made a note in the code but I think we should apply the `-b` bonus to the `initscore` calculation if set.

Also, it would be good to add tests for this. You could add something to [tests/e2e/cogs/initTracker_test.py](https://github.com/avrae/avrae/blob/e31cc9ed282d8cfa3052c7ddb391bec380b253f4/tests/e2e/cogs/initTracker_test.py), similar to the existing `TestYourStandardInitiative` class.

codybanman (approved):

PR #2227, opened 2025-11-11
fix: cross-shard DM interaction support

7he4lph4:
### Summary
Implements cross-shard IPC with O(1) targeted Redis pubsub

### Changelog Entry
Fixed `!i madd` selection buttons

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/utils/redisIO.py
+++ b/utils/redisIO.py
@@ -43,6 +43,11 @@ class RedisIO:
     async def delete(self, *keys):
         return await self._db.delete(*keys)
 
+    async def getdel(self, key, default=None):
+        """Atomically get and delete a key."""
+        encoded_data = await self._db.getdel(key)
+        return encoded_data.decode() if encoded_data is not None else default
+
     async def setex(self, key, value, expiration):
         return await self._db.setex(key, expiration, value)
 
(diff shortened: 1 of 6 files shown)


utils/selection/selection_monster.py
```diff
@@ -105,9 +105,7 @@ def create_embed(page: int) -> disnake.Embed:
     if interaction is not None:
         try:
             combat_channel = ctx.channel.mention
-            await interaction.followup.send(
-                f"✅ Adding **{key(choice)}** to combat in {combat_channel}!", ephemeral=True
-            )
+            await interaction.followup.send(f"✅ Adding **{key(choice)}** to combat in {combat_channel}!")
```
1drturtle, line 108: I don't think any other avrae messages contain the emojis

utils/redisIO.py
```diff
@@ -43,6 +43,11 @@ async def exists(self, *keys):
     async def delete(self, *keys):
         return await self._db.delete(*keys)
 
+    async def getdel(self, key, default=None):
+        """Atomically get and delete a key (Redis 6.2+)."""
```
1drturtle: probably don't need the version since we're already at `redis:7`

1drturtle (approved):

utils/redisIO.py
7he4lph4: I mean the `!init opt` cmd do, also thought it was a good visual feedback tick letting the user know immediately that their btn click has worked correctly

PR #2246, opened 2026-01-16
fix: prevent passive_effects dmg bonus from reducing healing

7he4lph4:
### Summary                                                                                                 
  - `passive_effects` damage bonuses now filter by type: bonuses with 'heal' in the string apply to healing rolls, 
  others apply to damage rolls                                                                                     
  - Fixes issue where `+2 [fire]` from an initiative effect would reduce `-1d8 [healing]` output                   
  - User-provided `-d` arguments are unaffected                                                                    
  - Damage penalties (e.g., `-2`) still work on damage rolls

### How to Test:
- Replace <char_name> with a test character to test this:
```
!multiline
!i begin
!i join
!action add "Fire Strike" -b 5 -d "1d8 [fire]"
!action add "Healing Touch" -d "-1d8 [healing]"
!i effect <char_name> "Damage Buff" -d "2 [fire]"
!a "Fire Strike" hit
!a "Healing Touch" hit
!i effect <char_name> "Healing Buff" -d "-3 [magical healing]"
!a "Fire Strike" hit
!a "Healing Touch" hit
!action delete "Fire Strike"
!action delete "Healing Touch"
```

### Behaviour Before:
image

### Behaviour After:
image

### Changelog Entry
Fixed damage bonuses from passive effects reducing healing output

#### PR Type

- [x] This PR fixes an issue.
#### Other

### DIFF
--- a/cogs5e/models/automation/effects/damage.py
+++ b/cogs5e/models/automation/effects/damage.py
@@ -83,7 +83,7 @@ class Damage(Effect):
         d_args = []
         if not (self.contains_roll_meta(autoctx) or self.fixedValue):
             d_args = args.get("d", [], ephem=True)
-            # #224: filter bonuses by sign: positive for damage, negative for healing
+            # #224: filter bonuses by type: 'heal' for healing, else for damage
             effect_bonuses = autoctx.caster_active_effects(mapper=lambda e: e.effects.damage_bonus, default=[])
             d_args.extend(utils.filter_dmg_bonuses(damage, effect_bonuses))
 
--- a/cogs5e/models/automation/effects/damage.py
+++ b/cogs5e/models/automation/effects/damage.py
@@ -78,15 +78,15 @@ class Damage(Effect):
         if autoctx.target.is_simple and self.is_meta(autoctx):
             return
 
-        d_args = []
-        # check if we actually need to care about the -d tag
-        if not (self.contains_roll_meta(autoctx) or self.fixedValue):
-            d_args = args.get("d", [], ephem=True)
-            # add on combatant damage effects (#224)
-            d_args.extend(autoctx.caster_active_effects(mapper=lambda effect: effect.effects.damage_bonus, default=[]))
-
         # set up damage AST
         damage = autoctx.parse_annostr(damage)
+        d_args = []
+        if not (self.contains_roll_meta(autoctx) or self.fixedValue):
+            d_args = args.get("d", [], ephem=True)
+            # #224: filter bonuses by sign: positive for damage, negative for healing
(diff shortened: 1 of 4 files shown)


DemiplaneJoseph:
What if something is reducing the damage dealt?

7he4lph4:
> What if something is reducing the damage dealt?

Valid. I will check if 'heal' is in the bonus string instead.

vaguely-happy:
> > What if something is reducing the damage dealt?
> 
> Valid. I will check if 'heal' is in the bonus string instead.

I think there are some aliases floating around that do "Damage Reduction" or similar - a quick test shows that Monk Deflect Attack action uses "reduction" as its type. I think for backward compatability I would like to see reduction added to the same category as healing and an additional test case for it.

(Great bug by the way - I've wanted this one for ages)

vaguely-happy:
> > What if something is reducing the damage dealt?
> 
> Valid. I will check if 'heal' is in the bonus string instead.

Not sure if you have got this now but I don't see it in the code at a glance. If not then perhaps something like this

keywords = ("heal", "reduction")
return [
    b for b in bonuses
    if is_healing == any(k in b.lower() for k in keywords)
]

7he4lph4:
> > > What if something is reducing the damage dealt?
> > 
> > 
> > Valid. I will check if 'heal' is in the bonus string instead.
> 
> Not sure if you have got this now but I don't see it in the code at a glance. If not then perhaps something like this
> 
> keywords = ("heal", "reduction") return [ b for b in bonuses if is_healing == any(k in b.lower() for k in keywords) ]

Ok yea, I think I understand now, so we need to treat 'reduction' as a healing dmg type. I don't think that's advisable since passive_effect heal bonuses will start adding to the defect attack action. Cross contamination between healing buffs and damage reductions.

This might have to corrected in the game data itself by Avrae devs. They can likely flip the sign to - in the game data. Something like: `-{dexterityMod} [reduction]`

Lathaon:
There are cases where other words are used. For example, the _Shield_ spell uses [shielded]

Personally, I think rather than hard-coding certain damage types to work differently, it would be better to have separate Damage and Healing nodes and make it so neither can go past 0. That would be a very big and breaking change, though...

7he4lph4:
yea, that would be an org level change, a slow rollout of v1 -> v2 maybe, but way out of scope of OSS contribs

VeryGreatFrog:
I believe this is a potentially bad idea. A better idea is to add a Healing node, which completely ignores damage effects, while Damage nodes continue to behave as they have in the past (incl negative damage which would be affected by damage effects).

Adding a healing node would not be difficult.

PR #2247, opened 2026-02-01
feat: add `reply_to_id` to `AliasCtx` and `-reply` option to `embed`

andresperezl:
### Summary
This include a new field `AliasCtx.reply_to_id` which is the message ID of the message this alias invocation is replying to, if any.

Additionally, it adds the `-reply` to the `embed` command, the value being the message ID of another message in the channel, which can be used for the generated embed to be sent as a response to said message.

This allows creating an alias like:
```
!alias smart_reply embed -title "Response" -desc "Acknowledged!" {{f"-reply {ctx.reply_to_id}" if ctx.reply_to_id else ""}}
```

image

In the image above the second message was not sent as a reply, so it doesnt link to any message

### Changelog Entry
add `reply_to_id` to `AliasCtx` and `-reply` option to `embed`

#### PR Type

- [ ] This PR fixes an issue.
#### Other

### DIFF
--- a/aliasing/api/context.py
+++ b/aliasing/api/context.py
@@ -19,6 +19,7 @@ class AliasContext:
         self._prefix = ctx.prefix
         self._alias = ctx.invoked_with
         self._message_id = ctx.message.id
+        self._reply_to_id = ctx.message.reference.message_id if ctx.message.reference else None
 
     @property
     def guild(self):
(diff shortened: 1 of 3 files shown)


andresperezl:
I have an actual use for this feature and that is that I have a simple alias that allows me and my friends, who have been playing for a few years via Avrae in Discord, to reply always in character, either actions or speech, via:

```
!alias cdo embed -thumb {{image}} -title "{{name}}" -color "{{color}}" -desc %*%
```

Because of the nature of PbP sometimes there are a lot of things happening, and you want to reply to stuff out of order, so you need to jump in OOC and tell what you were replying to, and it will be a lot easier if I could expand the command with the message that we were intended to respond to:

```
!alias cdo embed -thumb {{image}} -title "{{name}}" -color "{{color}}" {{f"-reply {ctx.reply_to_id}" if ctx.reply_to_id else ""}} -desc %*% 
```

I can see other uses for the reply option as a "recall" for information in some way, where the DM or Player can point to a conversation or an event that happened in the past and directly link it to their message

andresperezl:
There is already a small improvement that could be added but wanted to keep this PR small, and that is the ability to respond to message cross-channel, right now by the way the option is setup it only allows responding to messages in the same channel as the command/alias was sent.

Also, I thought about adding this to `echo` and `techo` but by the nature of those commands, I think remaining simple should be their objective.

Finally, right now if the message reference fails is simply ignored and it sents the message anyways (which could happen if the message id is passed manually to `-reply` and not fed directly from `AliasCtx.reply_to_id`, and the reply wont trigger a notification as I didnt want the `-reply` option to become a method of abuse, but there is also an argument to add an additional `-mention` option that would trigger the notification.

PR #2251, opened 2026-02-25
Changed authentication to v2

tonip-wizard:
### Summary
Upgraded the current auth service to v2 for most of the app, the v1 will still be used for character service as they don't yet support v2

### DIFF
--- a/ddb/client.py
+++ b/ddb/client.py
@@ -239,6 +239,10 @@ class BeyondClient(BeyondClientBase):
                 except (aiohttp.ContentTypeError, ValueError, TypeError):
                     log.warning(f"Cannot deserialize Auth Service response: {resp.status}: {await resp.text()}")
                     raise AuthException("Could not deserialize D&D Beyond response.")
+                token_val = data.get("token")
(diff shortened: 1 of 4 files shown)


ddb/client.py
```diff
@@ -214,18 +216,22 @@ async def _get_entity_entitlements(self, ctx, entity_type):
...
+        :param bool auth_v1: Whether to use the v1 auth endpoint, otherwise uses v2 as default.
         :returns: A tuple representing the short-term token for the user and its TTL, or (None, None).
         :rtype: tuple[str, int] or tuple[None, None]
         """
         body = {"Token": claim}
         try:
-            async with self.http.post(AUTH_DISCORD, json=body) as resp:
-                if not 199 < resp.status < 300:
+            async with self.http.post(AUTH_DISCORD_V1 if auth_v1 else AUTH_DISCORD_V2, json=body) as resp:
+                if resp.status == 401:
```
devgr, line 231: Just make sure that the code is correct for V1 error handling: V1 will return 200 status code, but the token will be `None`.

devgr (approved):

PR #93, opened 2018-08-07
Adds a `set_gvar(name,value)` to scripting.

Croebh:
Returns `None` if not found, `True` if edited, and `False` if no editor permission

PR #2271, opened 2026-06-30
feat(aliasing): add set_gvar/create_gvar to scripting

tiberius0954:
### Summary
Re-opening #93 

Adds the ability for aliases/scripts to write global variables (gvars), which were previously
read-only in scripting (`get_gvar`). Two new Draconic builtins are exposed:

- `set_gvar(address, value)` — edits an existing gvar's value.
- `create_gvar(value, script_writable=False)` — creates a new gvar owned by the invoking user and
  returns its address (UUID).

**NOTES**:
- **Opt-in, off by default.** A new `script_writable` field gates scripted writes, preserving the
  "gvars are static data" intent and protecting `using()` code libraries from being mutated at
  runtime. Owners toggle it with the new `!gvar scripting <address>` command; `create_gvar` can opt in
  at creation via `script_writable=True`. A missing field is treated as `False`, so **no migration is
  needed** and all existing gvars stay non-script-writable.
- **Permission preserved.** A script may only write a gvar if the *invoking user* (`ctx.author`) is
  its owner or an editor — the same check the `!gvar edit` command uses.
- **Deferred batch-commit**, mirroring `set_uvar`: writes are committed once at the end of execution,
  so repeatedly writing the same gvar in a loop results in a single DB write.
- **Abuse cap.** A single execution may create/edit at most 50 distinct gvars (charged before the
  lookup, so failed writes can't spam the DB either). Reaching the cap stops further writes but does
  **not** roll back writes already made — consistent with how uvars/character commits behave.
- Editor management, deletion, and toggling `script_writable` remain **owner-only Discord commands**
  (not exposed to scripting), since they widen the write trust boundary.

Docs (`docs/aliasing/api.rst`) and e2e tests (`tests/e2e/aliasing_mixed_test.py`) updated accordingly.

A corresponding avrae.io PR will be pushed shortly to reflect these new changes

### Changelog Entry
Aliasing: added `set_gvar` and `create_gvar` so scripts can write global variables, opt-in per gvar via `!gvar scripting`.

#### PR Type

- [ ] This PR fixes an issue.
#### Other


aliasing/evaluators.py
```diff
@@ -287,6 +298,67 @@ def get_gvar(self, address):
...
+        :param str value: The new value.
+        """
+        address = str(address)
+        value = str(value)
+        # validate (and account for the write) only the first time an address is dirtied this run
+        first_time = address not in self.gvars_changed and address not in self.gvars_created
+        if first_time:
+            # charge before the lookup so a caught-exception loop on a bad address can't spam reads
+            self._charge_gvar_write()
+            gvar = self.ctx.bot.mdb.gvars.delegate.find_one({"key": address})
```
1drturtle, line 325: isn't this async?

tiberius0954: `.delegate` is Motor's underlying sync PyMongo collection, so this isn't a coroutine. This matches the existing `get_gvar`/`get_svar`.

1drturtle (approved):
this is awesome btw!

PR #2213, opened 2025-09-01
feat: add option to hide ieffect

7he4lph4:
### Summary
- only hides effect description and parenthetical
- shows duration, conc, and name for combat context
- also adds automation docs about the new hidden bool for ieffects

image

### Changelog Entry
added option to hide user-added init effect's description and modifiers

#### PR Type

- [ ] This PR is a code change that implements a feature request.
- [ ] This PR fixes an issue.
- [x] This PR adds a new feature that is not an open feature request.
- [ ] This PR is not a code change (e.g. documentation, README, ...)
#### Other
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] If code changes were made then they have been tested.
- [ ] I have updated the documentation to reflect the changes.

PR #2290, opened 2026-08-06
Fix regressions with init descriptions

VeryGreatFrog:
### Summary
Fixes a regression introduced with #2213 where !i status messages would not send their truncate descriptions when they are too long.
Fixes a second regression where init effect description would display in !a when they shouldn't (introduced by #2213)

### Changelog Entry

#### PR Type

- [X] This PR fixes an issue.
#### Other


VeryGreatFrog:
Please review/test my changes as well. Test case: "`!i effect MyTarget MyEffect -desc "text that is 2000 characters long"`. On live, this will fail. This change should have fixed it and otherwise behave as expected as it looks for explicit description hidden kwargs.

PR #2300, opened 2026-09-17
Add Gear field for monsters

VeryGreatFrog:
### Summary
This supports Monsters to contain data about their Gear in their statblocks. This is a field that WOTC has been using more and more in recent statblocks.

#### PR Type

- [ ] This PR fixes an issue.
#### Other

image


VeryGreatFrog:
This adds support for both homebrew and official monsters. Since this data already exists on beyond, it should be easy to adjust the data ingestion process to also grab this data. Here is a list of all 148 official monsters with gear (as far as I know)
[creatures_with_gear.json](https://github.com/user-attachments/files/32356085/creatures_with_gear.json). Ideally, these would have their data updated.

Finally, I have tested these changes but I would appreciate a second look as this is my first time making changes in these classes.

